PR #44, opened 2026-04-22
#43 add icmp

### LINKED ISSUE #43: Allow ICMP protocol rules
Allow ICMP protocol rules to be created. Its appears that GitHub runners will sometimes attempt to ping the wire server. This will cause block events to appear in CargoWall logs, while this does not appear to negatively affect workflow runs - it does cause unnecessary noise. So ICMP should be default allowed on the azure wire server ip `168.63.129.16` from `cargowall-action`

matthewdevenny:
Closes #43 

- Adds option to allow ICMP
- Expands ALL to include ICMP (will be called out in release notes) 
- Adds ICMP allow on Azure wire server


jordanjennings (approved):
Changes LGTM

mtmk:
ICMP rules on IPv6 CIDRs can silently break TCP/UDP reachability.

`addCIDRv6` (firewall.go:283) keys `PortSpecific=1` off `len(ports) > 0`. The v6 BPF path unconditionally allows ICMPv6 (tcbpf.c:471) and never queries the port map with `proto=1`, so the ICMP port entry is dead, but the `PortSpecific=1` flip on the LPM entry still takes effect.

Failure cases under `default: deny`:

1. Same-CIDR rule collision. Rule A `{cidr: 2001:db8::/64, action: allow}` (all ports, PortSpecific=0) followed by rule B `{cidr: 2001:db8::/64, protocol: icmp, action: allow}` (PortSpecific=1). The second LPM update overwrites the first. Port map has only the dead ICMP entry. All TCP/UDP to that subnet is silently blocked.

2. Narrower ICMP rule shadowing a broader allow. Rule A `{cidr: 2001:db8::/48, action: allow}` plus rule B `{cidr: 2001:db8:1::/64, protocol: icmp, action: allow}`. LPM picks the /64, PortSpecific=1, TCP/UDP to that /64 denied.

Suggest rejecting `ProtocolICMP` on IPv6 CIDR rules at config load, alongside the existing port=0 check in `LoadConfig` and `LoadConfigFromCargoWall`:

```go
if p.Protocol == ProtocolICMP {
    if p.Port != 0 {
        return fmt.Errorf(\"ICMP rules must have port=0, got %d\", p.Port)
    }
    if rule.Type == RuleTypeCIDR {
        if ip, _, err := net.ParseCIDR(rule.Value); err == nil && ip.To4() == nil {
            return fmt.Errorf(\"ICMP protocol not valid on IPv6 CIDR %q; ICMPv6 is always allowed\", rule.Value)
        }
    }
}
```

Hostname rules resolving to v6 addresses are a separate, lower-severity case worth handling at the v6 resolution path.

PR #55, opened 2026-05-14
Add GitLab CI Support

matthewdevenny:
## Code changes

- `pkg/config/config.go` — added `AutoAddedTypeGCPInfrastructure` and `AutoAddedTypeGitLabService` constants; `EnsureInfraAllowed` now takes the auto-added-type as a parameter so callers can tag GCP vs Azure rules correctly.
- `cmd/flags.go` — added 7 orthogonal flags (`--dns-redirect-iptables`, `--docker-dns-interception`, `--dns-query-filtering`, `--prepopulate-dns-cache`, `--auto-allow-cloud-metadata`, `--auto-allow-github-hosts`, `--auto-allow-gitlab-hosts`), the `--gitlab-ci` preset, `--pidfile`, and `--ready-file` (path-shared with `wait-ready` via the same env var). `AfterApply` switches on `CIMode()` and expands the matching preset into the orthogonal flags — the two presets are mutually exclusive (GitHub wins if both are set). `WaitReadyCmd` and `StopCmd` are registered in the CLI struct.
- `cmd/start.go` — replaced all 11 `cmd.GithubAction` references with the appropriate orthogonal flag check; renamed `loadGitHubActionsConfig` → `loadCIConfig`; split `autoAllowInfraHosts` into `autoAllowCloudMetadata` (auto-detects Azure vs GCP via systemd-resolved upstreams) + `autoAllowGitHubHosts` + `autoAllowGitlabHosts` (new) + `autoAllowCodeCargoAPI`, plus a shared `autoAllowFromEnvURLs` helper. Pidfile is written **before** the ready sentinel (so `wait-ready` returning implies the pidfile exists) and pidfile-write failure is now **fatal** — `--pidfile` is an opt-in promise that `cargowall stop` will work, so silently best-effort would surprise the user.
- `cmd/stop.go` — extracted `writePidfile` helper (paired with `readPidfile`) using `os.OpenFile` with `O_NOFOLLOW` so a pre-planted symlink at the pidfile path can't be silently followed and clobbered. Also extracted the pidfile-removal helper to dedupe `stopProcess`.
- `cmd/wait_ready.go` — new subcommand that polls the sentinel with a timeout. Shares `--ready-file` / `CARGOWALL_READY_FILE` defaults with `start` so the two never disagree on the path.
- `cmd/summary.go` — doc note on `mapAutoAllowedType` flagging that the new `gcp_infrastructure` / `gitlab_service` types aren't yet in the proto enum, so events tagged with them have their `AutoAllowedType` field omitted from the SaaS push (follow-up to extend the proto).

## Tests

- `cmd/flags_test.go` — preset expansion (GitHub, GitLab, none, explicit-flag-preserved, both-set-precedence).
- `cmd/start_test.go` — `autoAllowGitlabHosts` defaults / env discovery / override; `autoAllowGitHubHosts` runtime URL discovery; `applyAutoAllowHelpers` dispatcher gating; `applyCloudMetadataAllows` GCP-only / Azure-detected / Azure host env override (pinning down the first-match-wins port-53 semantics); `autoAllowFromEnvURLs` malformed-URL skip.
- `cmd/wait_ready_test.go`, `cmd/stop_test.go` — already-exists / appears-before-timeout / times-out for `wait-ready`; pidfile parsing, write/read round-trip, **`O_NOFOLLOW` symlink-rejection** (pins the security property so a future refactor can't quietly drop the flag), SIGTERM-on-real-child + already-gone for `stop`.

## CI + docs

- `.github/workflows/ci.yml` — new `test-cargowall-gitlab-ci` job that exercises `--gitlab-ci`, `wait-ready`, and `stop` end-to-end. Uses a GitHub-hosted privileged Linux runner as the closest available stand-in for GitLab SaaS's privileged-container environment — see post-merge checklist below for real-environment validation.
- `README.md` — added a GitLab CI subsection with a copy-pasteable `.gitlab-ci.yml`, expanded the flag table with the 10 new flags, noted that `CARGOWALL_PIDFILE` is read by both `start` and `stop`, and shrank "What's not in the standalone path" since most pieces are now flag-accessible.

## Backwards compatibility

`--github-action` is preserved as a preset that expands to the same orthogonal flags, so `code-cargo/cargowall-action` keeps working unchanged. No coordinated release needed.

## Post-merge checklist

- [ ] Run a smoke `.gitlab-ci.yml` against gitlab.com SaaS shared runners with `--gitlab-ci --audit-mode` and confirm the audit log artifact looks sane (the in-PR CI job runs on a GH-Actions privileged-Linux stand-in, not the real GitLab SaaS environment).
- [ ] Open follow-up to extend the `CargoWallAutoAllowedType` proto with `gcp_infrastructure` / `gitlab_service` so those auto-allow tags survive the SaaS summary push.
- [ ] Open follow-up for the standard pidfile pid-reuse gotcha: if cargowall crashes (SIGKILL, OOM) without removing the pidfile and the kernel recycles the pid before `cargowall stop` runs, the unrelated process gets the SIGTERM. Mitigation: embed start time or `/proc/<pid>/comm` in the pidfile and validate before signaling.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -667,6 +667,126 @@ jobs:
           fi
           echo "Emergency cleanup complete"
 
+  test-cargowall-gitlab-ci:
+    name: Integration Test (GitLab CI mode)
+    needs: build-binary
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      actions: read
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Download binary
+        uses: actions/download-artifact@v8
+        with:
+          name: cargowall-binary
+          path: bin
+
+      - name: Make binary executable
+        run: chmod +x bin/cargowall
+
+      - name: Write minimal policy
+        run: |
+          sudo mkdir -p /etc/cargowall
+          sudo tee /etc/cargowall/config.json > /dev/null <<'JSON'
+          {
+            "defaultAction": "deny",
+            "rules": [
+              {
+                "type": "hostname",
+                "value": "github.com",
+                "ports": [{"port": 443, "protocol": "tcp"}],
+                "action": "allow"
+              },
+              {
+                "type": "cidr",
+                "value": "8.8.8.8/32",
+                "ports": [{"port": 53, "protocol": "udp"}],
+                "action": "allow"
+              }
+            ]
+          }
+          JSON
+
+      - name: Start CargoWall with --gitlab-ci preset
+        run: |
+          # nohup + & + disown lets cargowall outlive this step's shell.
+          # The privileged-Docker stand-in for GitLab SaaS doesn't exist on
+          # GitHub Actions runners, so we exercise the same code path on the
+          # Azure-backed Linux runner — eBPF + sudo + iptables work the same.
+          sudo nohup ./bin/cargowall start \
+            --gitlab-ci \
+            --audit-mode \
+            --audit-log /tmp/cargowall-audit.json \
+            --pidfile /tmp/cargowall.pid \
+            --dns-upstream "8.8.8.8:53" \
+            --debug \
+            > /tmp/cargowall.log 2>&1 &
+          disown
+
+      - name: Wait for ready via wait-ready subcommand
+        run: ./bin/cargowall wait-ready --timeout 30s
+
+      - name: Show startup log (debug aid)
+        if: always()
+        run: sudo cat /tmp/cargowall.log || true
+
+      - name: Verify pidfile written
+        run: |
+          if ! sudo test -s /tmp/cargowall.pid; then
+            echo "❌ pidfile /tmp/cargowall.pid was not written"
+            exit 1
+          fi
+          echo "✅ pidfile present, pid=$(sudo cat /tmp/cargowall.pid)"
+
+      - name: Allowed call reaches github.com
+        run: |
+          curl -sf --connect-timeout 10 --max-time 30 https://github.com > /dev/null
+          echo "✅ github.com reachable"
+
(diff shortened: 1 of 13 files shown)


jordanjennings (commented):
### Issues to consider before merge

1. `--ready-file` asymmetry — `WaitReadyCmd` exposes `--ready-file` (and `CARGOWALL_READY_FILE`) defaulting to `/tmp/cargowall-ready`, but `start` hardcodes the same path on the write side (cmd/start.go:437). A user who overrides the env var only on the wait-ready side will hang forever because start is still writing to /tmp/. Either add a matching `--ready-file` flag to StartCmd (parallels how `--pidfile` works) or drop the flag from wait-ready and document the path as fixed.

2. Pidfile open failure is silent (cmd/start.go:417-422) — if `O_NOFOLLOW` fires (symlink planted on a shared runner) or the parent dir doesn't exist, the user gets a Warn line and start continues. But then `cargowall stop --pidfile X` fails later with `read pidfile: no such file`. The whole reason the user passed `--pidfile` is so stop works; silent best-effort surprises them. Either return the error (fail-fast — they opted in) or bump to Error and say in the message that stop won't work.

3. No test for the `O_NOFOLLOW` symlink-guard — that flag is the *only* defense against the symlink-clobber attack, but nothing pins down the behavior. A table-test where the pidfile path is a symlink, asserting open fails and the target is untouched, would lock in the security property. Right now a future refactor could quietly drop the flag and tests would still pass.

### Nice-to-haves

- Pidfile pid-reuse race — `stopProcess` SIGTERMs whatever pid is in the file. If cargowall crashed and the OS recycled the pid before stop ran, an unrelated process gets the signal. Standard pidfile gotcha, not introduced here, but worth a future ticket — embed start time or `/proc/<pid>/comm` in the pidfile to validate.

### Verdict

Ship after addressing #1 and #2.

jordanjennings (approved):
LGTM

PR #57, opened 2026-05-15
#56 add dns search domains

### LINKED ISSUE #56: Add DNS Search Domains
When `--dns-query-filtering` is enabled, the DNS proxy refuses any name that doesn't match a hostname rule (anti-DNS-tunneling). This breaks a real use case: a user adds a VPC CIDR rule (e.g. `10.0.0.0/16`) to permit internal traffic, but DNS resolution for
 `ip-10-0-0-5.us-west-2.compute.internal` (AWS EC2 private DNS) is then refused — there is no hostname rule and the user doesn't want to write one because the names are ephemeral and uncountable.

Need to add configuration to allow DNS resolution and add generic search domain stripping

matthewdevenny:
Closes #56

## Summary

Adds configurable DNS search domains to the proxy and teaches `--auto-allow-cloud-metadata` to detect the host cloud (AWS / Azure / GCP) so the right provider-specific allows ship out of the box.

## New behavior

**`searchDomains` (config) / `CARGOWALL_SEARCH_DOMAINS` (env) / `search_domains` (proto)** — a list of DNS suffixes (e.g. `.compute.internal`) that:
- get stripped before hostname-rule matching, so a rule for `bastion` matches `bastion.compute.internal` (longest match wins, case-insensitive),
- and bypass DNS query filtering for unmatched names under the suffix, so cloud-internal names resolve without per-hostname tracking when the traffic is governed by a VPC CIDR rule.

Explicit `allow` / `deny` hostname rules still win over the bypass — a deny for `blocked.compute.internal` is honored even with `.compute.internal` configured. When both forms have rules, the narrower exact rule wins over the parent-suffix rule (both for deny and allow), so a `bastion` rule (port 22) correctly applies to `bastion.compute.internal` instead of being masked by a broad `compute.internal` allow. Cross-form mixed-action rules (e.g. `*.compute.internal: deny 80` + `bastion: allow 22`) now produce a `HostnameVerdict` that records both sides, so per-port BPF entries faithfully express both rules.

Suffixes must have ≥ 2 labels, follow RFC 1035/1123 label rules (no leading/trailing hyphen, ≤ 63 octets), and must not themselves be public suffixes per [Mozilla's PSL](https://publicsuffix.org/); `.com`, `.co.uk`, `.com.au`, `.github.io`, etc. are rejected at config load. Kubernetes suffixes (`.cluster.local`, `.svc.cluster.local`, `.default.svc.cluster.local`) are always active for stripping only and are rejected if a user tries to add them as user search domains (would inadvertently grant bypass). Matching is case-insensitive end-to-end (rule values are canonicalized at load, lookups lowercased).

**`--auto-allow-cloud-metadata` is now cloud-aware.** Detection signals, in order:
1. `CARGOWALL_CLOUD_PROVIDER=aws|azure|gcp` (case-insensitive; unknown values ignored)
2. DMI `sys_vendor` for AWS (`Amazon EC2`) / GCP (`Google*`)
3. DMI `chassis_asset_tag` for Azure (`7783-7084-3265-9085-8269-3286-77` — Azure-specific; avoids matching ordinary Hyper-V VMs the way `sys_vendor == "Microsoft Corporation"` would)
4. systemd-resolved upstreams containing `168.63.129.16` (Azure wireserver fallback for containers without DMI access)

Provider-specific allows added on top of the shared `169.254.169.254` IMDS allow: Azure wireserver + `.internal.cloudapp.net` + `trafficmanager.net` / `blob.core.windows.net`; AWS `.compute.internal` + `.ec2.internal`; GCP `.google.internal`.

## Proto additions

Two new `CargoWallAutoAllowedType` enum values are added — both additive, no consumer breaks by ignoring them:
- `CARGO_WALL_AUTO_ALLOWED_TYPE_CLOUD_METADATA = 5` — for the tag rename above; wired into `cmd/summary.go:mapAutoAllowedType`.
- `CARGO_WALL_AUTO_ALLOWED_TYPE_GITLAB_SERVICE = 6` — wires the previously-unmapped `gitlab_service` auto-added type (used by `autoAllowGitlabHosts` and the CI_* env var discovery) through to SaaS audit events.

## Dependency updates

- `github.com/cilium/ebpf` v0.20.0 → **v0.21.0** — verified safe for cargowall (no XDP usage, Linux-only build sidesteps Windows helper renames, no callers of removed APIs in our source or in bpf2go-generated output).
- `golang.org/x/net/publicsuffix` — promoted from transitive to direct dep; used to reject public-suffix search domains at config load.
- `github.com/alecthomas/kong` 1.13 → 1.15
- `golang.org/x/sys` 0.40 → 0.44
- `google.golang.org/genproto/googleapis/api` updated
- Plus transitive `x/mod`, `x/sync`, `x/tools`, `objx`, `go-spew`, `go-difflib` bumps

## Testing

- `go test ./pkg/config/ ./pkg/dns/ ./cmd/` — green in linux container
- `go vet ./...` — clean
- New CI integration steps in `test-cargowall-gitlab-ci` exercise the search-domain bypass end-to-end (dig under the configured suffix is NOT logged as `dns_blocked`; control query IS) and confirm the new cloud-detection log line is emitted
- BPF tests deferred to CI (`make test-bpf` needs root + CAP_BPF)
- Manual verification pending on a real GitLab SaaS runner (5.15 kernel) — separate from this PR, tracked alongside the legacy-TC-fallback follow-up


jordanjennings:
**Cross-form deny+allow with disjoint ports: the allow silently drops.**

`MatchHostnameRule` unions ports only when both forms deny (via `mergeDenyPorts`). With rules like:

- `*.compute.internal: deny 80`
- `bastion: allow 22`

a query for `bastion.compute.internal` returns deny-on-80; port 22 falls through to the default action. The PR description says "narrower exact rule wins over the parent-suffix rule," which holds within same-action but not across mixed-action — a user writing these two rules would probably expect SSH to bastion to work.

jordanjennings (approved):
LGTM

PR #61, opened 2026-05-29
#58 add clsact fallback

### LINKED ISSUE #58: Add legacy TC (clsact) fallback so cargowall works on kernel 5.8+
cargowall's TC egress attach uses `link.AttachTCX` (`pkg/network/network.go:63`), which requires the TCX hook merged in **Linux 6.6** (October 2023). However:

- `README.md:159` advertises support for "Linux kernel **5.x or newer**".
- `pkg/ebpf/detect.go:99` (`isKernelSupported`) returns `true` for any 5.x kernel.
- The `--gitlab-ci` flag and `README.md:233-263` explicitly target GitLab CI, but **GitLab SaaS shared runners run Linux 5.15.x** — below the TCX floor.
- `README.md:235` hedges with "run a smoke job first to confirm `cargowall start --gitlab-ci` actually attaches in your project's runner image." That caveat papers over the fact that it almost certainly *doesn't* attach today.

On a 5.x or 6.0–6.5 kernel, `link.AttachTCX` returns a kernel error (`EINVAL` / `EOPNOTSUPP`) and cargowall fails to start, rather than falling back to the classic attach path that worked on those kernels for years.

matthewdevenny:
Closes #58 

  ## What & why

  cargowall attached its egress eBPF program with `link.AttachTCX`, which requires the **TCX hook merged in Linux 6.6**. On older kernels the attach failed and cargowall refused to start — even though the classic **`clsact`** attach path (Linux 4.5+) worked there for years. This contradicted
  the documented "5.x or newer" support claim and, in particular, broke **GitLab SaaS shared runners (Linux 5.15)**, where `cargowall start --gitlab-ci` could not attach.

  This PR makes the egress attach **prefer TCX on 6.6+ and transparently fall back to a legacy `clsact` qdisc + direct-action `cls_bpf` filter on 5.8–6.5**, so cargowall works on kernel **5.8+**. (5.8 is the real floor: `bpf/tcbpf.c` uses `BPF_MAP_TYPE_RINGBUF`, which landed in 5.8 — the
  attach mechanism was never the limiter.)

  ## Changes

  - **New `pkg/tc` package** owns the egress attach. `tc.AttachEgress` tries `link.AttachTCX` and, **only** on `ebpf.ErrNotSupported`, falls back to `clsact`; any other error is fatal (fail-closed). It returns an `io.Closer`, so the caller's teardown is mechanism-agnostic. `AttachTC` is
  removed from `pkg/network`.
  - **Hand-rolled netlink clsact** (`pkg/tc/clsact_linux.go`) built on `github.com/mdlayher/netlink` — already a transitive dependency via `cilium/ebpf`, so **no new module tree** is introduced. It creates the clsact qdisc only if absent (libbpf-style ownership: teardown removes the qdisc
  only if we created it) and attaches the program as a **direct-action** filter.
    - ⚠️ Direct-action (`TCA_BPF_FLAG_ACT_DIRECT`) is mandatory: without it `cls_bpf` treats the program's return code as a classid rather than a verdict, so `TC_ACT_SHOT` wouldn't drop — a silent fail-open. It is set and unit-tested.
  - **Kernel floor corrected to 5.8** in `pkg/ebpf/detect.go` (`isKernelSupported` previously accepted any 5.x) plus clearer messaging; README requirements + the CI-mode error strings are updated, and the GitLab section now states SaaS 5.15 attaches via clsact automatically.
  - **No BPF changes** — the same `SCHED_CLS` object attaches via either mechanism (audited: no CO-RE relocations; ringbuf at 5.8 is the newest feature it uses).
  - Bumps `cargowall-action` to **v1.3.0** across the workflows + README, and adds `iproute2` to the CI test job so the integration test's `tc filter show` assertions run in CI.

  ## Testing

  - **Unit (kernel-independent, runs in CI):** `isKernelSupported` table (4.19/5.4/5.7 rejected; 5.8/5.15/6.6 accepted); `htons`/`makeFilterInfo` byte-order packing; `encodeEgressFilterAttrs` decodes the filter attrs and asserts the direct-action flag (the fail-open guard).
  - **Root-gated integration** (`./pkg/tc` added to the sudo test lane): attaches a minimal `SCHED_CLS` program to `lo` through the clsact path, verifies a `direct-action` filter via `tc filter show`, then detaches and confirms cleanup.
  - Verified locally: `go build` / `go vet` / `staticcheck` / `gofumpt` clean (`GOOS=linux`); full `./pkg/...` suite green in a privileged container with `iproute2`.

### DIFF
--- a/.github/workflows/bpf-generate.yml
+++ b/.github/workflows/bpf-generate.yml
@@ -17,7 +17,7 @@ jobs:
       contents: read
       id-token: write
     steps:
-      - uses: code-cargo/cargowall-action@fc9644606d84536b4e59a30ad2802904f7b4ff22 # v1.2.0
+      - uses: code-cargo/cargowall-action@41060a3cd1b22ea4922be4a8769c149983ef4eae # v1.3.0
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       actions: read
       id-token: write
     steps:
-      - uses: code-cargo/cargowall-action@fc9644606d84536b4e59a30ad2802904f7b4ff22 # v1.2.0
+      - uses: code-cargo/cargowall-action@41060a3cd1b22ea4922be4a8769c149983ef4eae # v1.3.0
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
@@ -56,7 +56,7 @@ jobs:
       id-token: write
       checks: write
     steps:
-      - uses: code-cargo/cargowall-action@fc9644606d84536b4e59a30ad2802904f7b4ff22 # v1.2.0
+      - uses: code-cargo/cargowall-action@41060a3cd1b22ea4922be4a8769c149983ef4eae # v1.3.0
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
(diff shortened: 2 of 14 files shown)


jordanjennings (approved):
LGTM, pending the additional testing on GitLab discussed separately

Issue #65, opened 2026-06-24
Inconsistent hostname case in reporting (DNS-query wire case vs connection-event lowercase)


matthewdevenny:
Hostname **matching** is uniformly case-insensitive, but the hostname **case shown in logs / audit output** differs depending on which subsystem emitted the record:

- **DNS proxy path** preserves the original wire case. `pkg/dns/server.go` (~L410–418) deliberately keeps the raw `fullHostname` for log attribution and only lowercases a separate `canonicalHostname` for matching (there's an explicit comment to this effect). All audit/debug calls in that path emit the wire-case form.
- **Connection-event path** (`pkg/events/events.go`) logs whatever the IP→hostname lookup returns, which is mostly **canonical lowercase**: forward DNS mappings are stored lowercase via `UpdateDNSMapping(canonicalHostname, …)`, and reverse-PTR names are lowercased to the rule value when they match a tracked rule.

Net effect: the same host can appear as `GitHub.com` on a DNS-query audit line and `github.com` on a connection-event line. Not a correctness/enforcement bug (matching is already case-insensitive), but a reporting inconsistency worth normalizing for cleaner audit logs.

### Possible fix
Pick one canonical case for reporting (most likely lowercase, to match the stored mappings) and apply it consistently across both paths — or deliberately preserve wire case everywhere. Decide and make the two paths agree.

_Found while implementing #52 (port union across duplicate rule entries)._

Issue #64, opened 2026-06-24
Canonicalize CIDR rule values so equivalent forms dedup and match consistently


matthewdevenny:
`normalizeRules` (`pkg/config/config.go`) lowercases hostname rule values but leaves **CIDR** rule values untouched. As a result, textually-different-but-equivalent CIDR values are treated as distinct entries:

- IPv6 case: `2001:DB8::/32` vs `2001:db8::/32`
- bare IP vs explicit prefix: `8.8.8.8` vs `8.8.8.8/32`
- non-canonical zero-compression: `2001:db8:0:0::/64` vs `2001:db8::/64`

This matters for the rule-dedup / port-union work in #52: that merge groups rules by `(Type, Value, Action)` on the literal `Value` string, so equivalent-but-differently-spelled CIDRs won't merge their ports (the union is keyed off the raw string today). It can also produce duplicate/confusing entries in audit output.

### Proposal
Canonicalize CIDR `Value` at load time in `normalizeRules` (parse via `net.ParseCIDR` / `net.ParseIP`, normalize bare IPs to `/32`/`/128`, then re-emit the canonical `IPNet.String()`), so `cm.config.Rules` and `cm.resolvedRules` carry one canonical form. Keeping it load-time-only preserves the existing "config.Rules and resolvedRules agree byte-for-byte" invariant that hostname lowercasing already relies on.

_Found while implementing #52 (port union across duplicate rule entries)._

PR #66, opened 2026-06-24
#52 merge common rule entries

### LINKED ISSUE #52: Multiple Hostname Rule Entries
If the policy contains multiple rules with the same hostname, it appears the last one on the list determines the allowed ports for that hostname. If there are multiple entries with the same hostname, cargowall should add them together rather than last one wins

matthewdevenny:
Closes #52

## Problem

When a policy lists the same hostname more than once, only one entry's allowed
ports took effect instead of the union — effectively "last/first one wins,"
silently dropping ports the operator configured.

Root cause: in `matchHostnameRuleLocked` each hostname rule is classified into a
single candidate slot, so a second rule for the same hostname overwrote (or was
discarded by) the first and its `Ports` were lost. The same class of bug
affected duplicate CIDR rules — `CheckIPRuleConflict` reads the raw
`cm.config.Rules` while firewall BPF writes read `cm.resolvedRules`, so split
ports across duplicate CIDRs weren't unioned either.

## Fix

A single, centralized merge step in the config layer:

- Added `mergeDuplicateRules`, called once in `applyLoadedConfig` (the shared
  tail of all four loaders) right after `normalizeRules`/`validateRules`. It
  collapses rules sharing `(Type, Value, Action)` into one rule whose `Ports`
  are the **union** of the duplicates'. Because both `cm.config.Rules` and
  `cm.resolvedRules` are built from this slice, every consumer — hostname
  matching, CIDR conflict detection, and firewall BPF writes — observes the same
  unioned port set, with no changes to the matching-precedence logic.
- Renamed `mergeDenyPorts` → `unionPorts` (action-neutral) and reused it for the
  union; the "empty Ports = all ports" sentinel is preserved (any all-ports
  duplicate makes the whole group all-ports).

### Scope decisions

- Union applies to **both hostnames and CIDRs**, keyed on the literal `Value`
  string.
- Union is scoped to a **matching action** — opposite `allow`/`deny` rules on
  the same value stay a conflict governed by the existing precedence logic
  (unchanged); you can't union an allow-set with a deny-set.

## Tests

Added 7 tests (driven through `LoadConfigFromRules`, which routes through the
merge): hostname allow-union, all-ports sentinel absorbing (both orders),
wildcard-pattern union, deny union, CIDR union visible to `CheckIPRuleConflict`,
opposite-action rules **not** merged, and a no-duplicates-unchanged guard.

`go test ./pkg/...`, `go vet ./...`, and `gofmt -l` all clean.

## Follow-ups (filed)

- #64 — Canonicalize CIDR rule values (IPv6 case, bare-IP vs `/32`,
  zero-compression) so equivalent forms dedup/match. This is why the merge keys
  on the literal string for now.
- #65 — Inconsistent hostname case in reporting (DNS-query wire case vs
  connection-event lowercase).

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -208,6 +208,7 @@ func (cm *Manager) applyLoadedConfig(cfg *FirewallConfig) error {
 	if err := validateRules(cfg.Rules); err != nil {
 		return err
 	}
+	cfg.Rules = mergeDuplicateRules(cfg.Rules)
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
 	cm.config = cfg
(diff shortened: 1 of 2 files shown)


jordanjennings (approved):
LGTM

PR #69, opened 2026-06-24
#64 canonicalize cidr rules

### LINKED ISSUE #64: Canonicalize CIDR rule values so equivalent forms dedup and match consistently

matthewdevenny:
Closes #64.

## Problem

`normalizeRules` lowercased hostname rule values but left **CIDR** values untouched, so textually-different-but-equivalent CIDRs were treated as distinct entries — they wouldn't dedup/merge ports in the #52 rule-merge (which keys on the literal `Value` string) and produced confusing duplicate audit output. Examples:

- IPv6 case: `2001:DB8::/32` vs `2001:db8::/32`
- bare IP vs explicit prefix: `8.8.8.8` vs `8.8.8.8/32`
- non-canonical zero-compression: `2001:db8:0:0::/64` vs `2001:db8::/64`

## Change

Canonicalize CIDR `Value` at load time in `normalizeRules`, alongside the existing hostname lowercasing. A new `canonicalCIDR` helper parses via `net.ParseCIDR` / `net.ParseIP`, promotes bare IPs to `/32` / `/128`, and re-emits the canonical `IPNet.String()`:

| input | output | reason |
|---|---|---|
| `2001:DB8::/32` | `2001:db8::/32` | IPv6 case-folded |
| `2001:db8:0:0::/64` | `2001:db8::/64` | zero-compression canonicalised |
| `8.8.8.8` | `8.8.8.8/32` | bare IPv4 → explicit /32 |
| `2001:db8::1` | `2001:db8::1/128` | bare IPv6 → explicit /128 |
| `10.0.0.5/8` | `10.0.0.0/8` | host bits masked to network |

Values that parse as neither are returned unchanged, so `resolveRules` still logs and skips them. This keeps the existing "`config.Rules` and `resolvedRules` agree byte-for-byte" invariant that hostname lowercasing already relies on, and runs once at load time so every downstream consumer (hostname matching, CIDR conflict detection, BPF writes, audit) sees one canonical form.

### Knock-on cleanup

Because every config-sourced CIDR value is now an explicit-prefix CIDR, the bare-IP `net.ParseIP` fallbacks in `CheckIPRuleConflict`, `hasCIDRRule`, and `hasCIDRRuleAllPorts` (which only ever read post-load canonical rules) are dead and were removed — a `/32` / `/128` host route matches its exact IP via `IPNet.Contains` identically. As a bonus this fixes a latent specificity bug where an IPv6 single-IP rule was hard-coded to rank as 32 bits; it now correctly ranks as 128. `resolveRulesLocked` keeps its single-IP fallback (now annotated) because the public `resolveRules()` test wrapper drives it with un-normalized rules.

## Behavior notes

- Audit strings for CIDR rules now show the canonical form (e.g. a conflicting rule reported as `10.0.0.5/32` rather than `10.0.0.5`, or `10.0.0.0/8` rather than `10.0.0.5/8`). This is the intended dedup/clarity improvement from #64.
- The default-route exact-string special-cases in `firewall.go` (`== "0.0.0.0/0"` / `== "::/0"`) are unaffected — those strings are already canonical — and now also catch non-canonical spellings of the wildcard route.
- **Stricter load validation (intended):** an ICMP-port rule on a *bare* IPv6 address (e.g. `2001:db8::1`) now fails load, because canonicalization to `/128` makes `isIPv6CIDR` true and `validateRules` rejects ICMP-on-IPv6 (proto 1 is IPv4-only). The explicit-`/128` form already failed; this makes the bare and explicit forms consistent.

## Testing

- New `TestCanonicalCIDR` (table-driven: all issue cases + unparseable/empty passthrough).
- New `TestMergeDuplicateRules_EquivalentCIDRFormsUnion` — three spellings of one IPv6 host collapse to a single rule with unioned ports.
- Updated `TestNormalizeRules_*` and the `CheckIPRuleConflict` single-IP case to expect the canonical form.
- `go build`, `go vet`, `gofumpt`, and `go test ./pkg/config/ ./pkg/dns/...` all green (run under `GOOS=linux` / Linux container; the package is `//go:build linux`). The only failing suite is the privileged eBPF tests (`bpf`, `pkg/tc`), which require root and are unrelated.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1279,15 +1279,14 @@ func (cm *Manager) CheckIPRuleConflict(ip net.IP, hostname string, hostnameActio
 			continue
 		}
 
-		// Parse CIDR
+		// Rule values are canonical explicit-prefix CIDRs — normalizeRules
+		// promotes bare IPs to /32 or /128 at load time — so ParseCIDR always
+		// succeeds. A parse failure means a malformed value that resolveRules
+		// already logged and skipped; skip it here too. A /32 or /128 host
+		// route matches only its exact IP via Contains, ranking most-specific
+		// at ones==32/128 (an IPv6 single IP correctly ranks as 128, not 32).
 		_, ipnet, err := net.ParseCIDR(cm.config.Rules[i].Value)
 		if err != nil {
-			// Try as single IP
-			if ruleIP := net.ParseIP(cm.config.Rules[i].Value); ruleIP != nil && ruleIP.Equal(ip) {
-				// Exact match is most specific (32 bits)
-				mostSpecificRule = &cm.config.Rules[i]
-				mostSpecificBits = 32
-			}
 			continue
 		}
 
@@ -1558,13 +1557,12 @@ func (cm *Manager) hasCIDRRuleAllPorts(ipStr string) bool {
 			continue
 		}
 
+		// Rule values are canonical explicit-prefix CIDRs (normalizeRules
+		// promotes bare IPs to /32 or /128 at load time), so ParseCIDR
+		// succeeds and host routes match via Contains; a parse failure is a
+		// malformed value resolveRules already skipped.
 		_, ipnet, err := net.ParseCIDR(rule.Value)
-		if err != nil {
-			ruleIP := net.ParseIP(rule.Value)
-			if ruleIP == nil || !ruleIP.Equal(ip) {
-				continue
-			}
-		} else if !ipnet.Contains(ip) {
+		if err != nil || !ipnet.Contains(ip) {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


mtmk (approved):
LGTM

PR #70, opened 2026-06-24
#65 normalize hostname case to lowercase in reporting

### LINKED ISSUE #65: Inconsistent hostname case in reporting (DNS-query wire case vs connection-event lowercase)

matthewdevenny:
Closes #65.

## Problem

Hostname **matching** is uniformly case-insensitive, but the hostname **case shown in logs / audit output** differed by subsystem:

- **DNS proxy path** (`pkg/dns/server.go`) deliberately kept the raw wire-case `fullHostname` for attribution.
- **Connection-event path** (`pkg/events/events.go`) mostly reported the canonical lowercase form from the IP→hostname mapping.

Net effect: the same host could appear as `GitHub.com` on a DNS-query audit line and `github.com` on a connection-event line. Not an enforcement bug — purely a reporting inconsistency.

## Fix

Normalize on **lowercase** everywhere (matching the stored mappings):

- **DNS resolution intercept** — drop the separate wire-case `fullHostname`; use the lowercased `canonicalHostname` for all log / audit / `applyVerdictSide` attribution.
- **DNS query filtering** — lowercase `domain` so the block logs and the `LogDNSBlocked` audit record are lowercase. `isQueryAllowed` / `MatchHostnameRule` fold case internally, so matching behavior is unchanged.
- **Connection-event PTR path** — lowercase an unmatched PTR name (PTR replies can carry mixed / 0x20-randomized case off the wire). `FindTrackedHostname` and `UpdateDNSMapping` already lowercase internally, so this only closes the "mostly" gap.

## Testing

- New `TestHandleDNSQuery_BlockedQueryReportedLowercase`: a mixed-case blocked query (`Evil.Tunnel.Example.COM.`) is recorded as `evil.tunnel.example.com` in the audit log.
- `go test ./pkg/dns/... ./pkg/events/... ./pkg/config/...` pass on linux (run in a `golang:1.25` container). The eBPF tests (`bpf`, `pkg/tc`) require privileged kernel access and are unaffected by this change.

### DIFF
--- a/pkg/dns/server.go
+++ b/pkg/dns/server.go
@@ -320,7 +320,10 @@ func (s *Server) handleDNSQuery(w dns.ResponseWriter, r *dns.Msg) {
 	// DNS tunneling). isQueryAllowed handles both the full and the
 	// search-domain-stripped form internally with the right precedence.
 	if s.filterQueries && len(r.Question) > 0 {
-		domain := strings.TrimSuffix(r.Question[0].Name, ".")
+		// Lowercased for reporting consistency (#65): isQueryAllowed and
+		// MatchHostnameRule fold case internally, so this only affects the
+		// case shown in the block logs and the LogDNSBlocked audit record.
+		domain := strings.ToLower(strings.TrimSuffix(r.Question[0].Name, "."))
 		if !s.isQueryAllowed(domain, r.Question[0].Qtype) {
 			// Check if we're in audit mode - log but don't block
 			isAuditMode := s.auditLogger != nil && s.auditLogger.IsAuditMode()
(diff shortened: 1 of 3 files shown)


mtmk (approved):
LGTM

Issue #67, opened 2026-06-24
Derived CNAME-target allow is query-filter-only — direct target queries return un-enforced IPs (CDN-variant split)


matthewdevenny:
## Background

#62 added transparent CNAME support, but scoped it to **DNS query filtering only**: when an allowed host's response carries a CNAME chain, the target *names* are learned (`cnameAllowed` LRU in `pkg/dns/server.go`) so a later direct query for a target is forwarded instead of `REFUSED`. The IP/enforcement side was assumed already-handled because "a single forwarded response already carries every hop of the chain" — `extractIPsFromResponse` pulls the terminal A/AAAA records and they're allowed under the query name's rule.

That assumption holds when the client resolves the origin and consumes the whole chain in one response. It **breaks when the chain is split across multiple query round-trips**, which is exactly what CDN-fronted PKI endpoints do.

## The gap

`cnameAllowed` only un-`REFUSE`s the DNS *query*. It carries no ports and is never consulted by enforcement. So when a client issues a **direct query for a derived-allowed CNAME target**:

1. The query is forwarded (good) and returns real A/AAAA records.
2. `MatchHostnameRule(target)` matches no rule, so the switch in `handleDNSQuery` hits the `!verdict.Matched()` branch — "tracked (no rules yet)" — and **writes nothing to the BPF allow map**.
3. The subsequent connection to those IPs is **blocked**, unless those exact IPs were already allowed via the origin's in-band resolution.

This is two reported symptoms with one root cause:

- **(a) Direct-target-query enforcement gap.** A target un-refused by `cnameAllowed` returns IPs that never reach the allow map.
- **(b) CDN-variant split.** DigiCert/Symantec PKI splits between Akamai and Cloudflare per query/region. The proxy pre-resolves `s.symcd.com` to the Akamai variant (terminal `23.11.32.159`, allowed), but a fresh query for `mpki-ocsp.digicert.com` (matches `*.digicert.com`) can return the **Cloudflare** variant, whose terminal CNAME `mpki-ocsp.digicert.com.cdn.cloudflare.net` is learned but whose Cloudflare IPs are never enforced.

## Observed (GitHub Actions, Azure runner, enforce + DNS query filtering)

Denies on CNAME-target names of allowed hosts:

- `mpki-ocsp.digicert.com.cdn.cloudflare.net` — Cloudflare-variant terminal of allowed `s.symcd.com` / `*.digicert.com`.
[...]

Issue #62, opened 2026-06-22
CNAME-target names under DNS query filtering (IP path already transparent)


matthewdevenny:
## Original ask

Add support so that if an allowed host CNAMEs to another host it's transparently allowed, versus having to make entries for all of the potential CNAME redirects.

For example — you have an allowed rule for `builds.dotnet.microsoft.com` which might have a moving CNAME → `a441.dscd.akamai.net`, which would force you to wildcard on that since enumerating all of the possibilities would be whack-a-mole.

### IP-level CNAME following already works

Real recursive resolvers return the whole chain in a single response:

```
builds.dotnet.microsoft.com.  CNAME dotnetcli.trafficmanager.net.
                              CNAME builds.dotnet.microsoft.com.edgesuite.net.
                              CNAME a441.dscd.akamai.net.
a441.dscd.akamai.net.   A  23.56.109.145
a441.dscd.akamai.net.   A  23.56.109.139
```

cargowall already handles this:

- `extractIPsFromResponse` (`pkg/dns/server.go`) walks every A/AAAA in the answer section and ignores the owner name, so the Akamai A records are extracted even though they belong to `a441.dscd.akamai.net`.
- The verdict is computed against the **query name** (`MatchHostnameRule("builds.dotnet.microsoft.com")` → allow), and those IPs are written to the BPF allow map.
- Enforcement is purely IP-based.

So the app connecting to the Akamai IPs is **already allowed under the original rule — no `*.akamai.net` wildcard is needed** for the IP path. Short TTLs just re-trigger the same path on each rotation. Covered by the existing test "CNAME chain followed by A — only A extracted" in `pkg/dns/server_test.go`.

### The actual remaining gap: DNS query filtering refuses CNAME-target *names*

`isQueryAllowed` gates queries on the query name vs. the rule set. With `--dns-query-filtering` enabled (off by default, but bundled into `--github-action` and `--gitlab-ci`), a **separate** query for a CNAME target (e.g. `a441.dscd.akamai.net`) matches no rule and gets `REFUSED`.

This only affects clients that chase CNAMEs themselves (issuing a distinct query for the target). Standard glibc/Go stub resolvers send one query and consume the full chain, so they're unaffected.

## Refined scope

Allow CNAME-target **names** to be queried under DNS query filtering when they were learned from a response to an already-allowed query.

Proposed approach (localized to `pkg/dns/server.go`):

[...]

PR #73, opened 2026-06-26
#72 CNAME attribution reporting

### LINKED ISSUE #72: CNAME Attribution Followup
#63 and #68 added support for transparently following CNAMEs from allowed hostname rules - but this introduces a reporting gap. Users will see that a CNAME hostname is allowed but not know why it was allowed. Change reporting so it logs the actual hostname that allowed the CNAME and provide the CNAME chain in the event details

matthewdevenny:
Closes #72

## Why

Transparent CNAME support (#62/#63, #67/#68) lets cargowall allow the resolved IPs of a CNAME **target** of an allowed host — even when the chain is split across separate DNS round-trips (CDN-fronted PKI, dynamic edge labels). That closed a real enforcement gap but created a **reporting** one: when a client queries a CNAME target directly, its IP maps to that opaque edge name (e.g. `e13678.dscb.akamaiedge.net`), so connection events look like traffic to a host the user never allow-listed, with no explanation.

This PR reports those events under the **enabling rule hostname** the user actually allowed (e.g. `www.microsoft.com`) and surfaces the full CNAME chain as a drill-down, plus adds a log when CNAME targets are learned so the allow path is auditable.

## What

- **Retain origin + chain (`pkg/dns/server.go`):** `cnameAllowed`'s value goes from `[]config.Port` to `derivedAllow{ports, chain}`, carrying the full ordered chain (`origin → … → target`). The chain is built while walking `cnameChainTargets`, and transitive (split-round-trip) learning extends the parent's chain. `mergeDerivedAllow` unions ports across origins while keeping the first chain (stable attribution).
- **IP → origin attribution (`pkg/config/config.go`):** `RecordCNAMEChain` / `LookupCNAMEChain` over a per-IP, bounded, **recency-ranked set of origins** (`ipToCNAMEOrigins`), populated in the derived-enforcement loop (same place the IPs are written to BPF) and sharing the existing `ipLastSeen` / `cleanupOldEntries` eviction. A single CDN/edge IP is reachable via several allowed origins; `mergeDerivedAllow` keeps the chain last-write-wins and `LookupCNAMEChain` returns the most recently-seen origin whose chain is still live ("currently chains here"), falling back to the most recent once all have expired — i.e. the origin the **current request** most likely chased, not just the first one learned.
- **Report under the origin (`pkg/events/events.go`):** before the late-allow check, swap the reported hostname to `chain[0]` and attach the chain. Placing the swap before late-allow also makes a *blocked* derived connection on a non-inherited port attribute to the origin and evaluate against the origin's rule.
- **`cname_chain` drill-down field:** added to live logs (`"cname_chain", origin → … → target`, only when present) and to the audit JSONL (`AuditEvent.CNAMETarget`, `json:"cname_chain,omitempty"`); threaded through the four `Log*` calls.
- **Cloud summary push:** new `repeated string cname_chain = 11` on `CargoWallActionEvent` (regenerated `pb`), mapped in `auditEventToProto`. The core attribution (`DstHostname` = origin) already flows to the summary via `event.Hostname` with no proto change.
- **Learn-time logging (`pkg/dns/server.go`):** `Learned CNAME target` at Info on first-learn (Debug on refresh), recording `target`, `origin`, full `chain`, `inherited_ports`, and `source` (`rule:<pattern>` vs `derived`). Makes "was this edge ever learned, and from which origin?" answerable from the logs.

## Out of scope / follow-ups

- Rendering `cname_chain` in the cloud **Baseline Entries** table lives in the Controller service (C# `Controller.ApiService`, see the proto `csharp_namespace`) — this PR only populates the field.
- Startup pre-population needs no change: Phase-2 resolves through the proxy (so chains are learned), and Phase-1 system-cache IPs already map directly to the configured hostname.

## Attribution accuracy

For an IP shared by multiple allowed origins, attribution uses the most recently-resolved origin that still chains to it (see above), which matches the origin a chain-chasing client is currently following. Residual: connection events are processed asynchronously off the ringbuf, so "most-recent live origin at report time" can still lag connect time — this narrows the window but doesn't fully close it (closing it would need connect-time capture, impractical given the systemd-resolved DNS indirection). Ports are unioned regardless, so **enforcement is unaffected** — this only refines the display/audit attribution, and the full `cname_chain` chain is always shown so the real edge is visible either way.

## Testing

- New/updated unit tests in `pkg/dns`, `pkg/config`, `pkg/events`: chain learning + transitive extension, `mergeDerivedAllow` first-chain/union, `RecordCNAMEChain`/`LookupCNAMEChain` round-trip + copy + eviction, and `processEvent` origin attribution + `cname_chain` on both the allowed and blocked paths.
- `GOOS=linux go build ./...`, `staticcheck ./...`, and `gofumpt -l` are clean. Unit tests run on linux/CI (the dev host is darwin; the eBPF packages are `//go:build linux`).

### DIFF
--- a/cmd/summary.go
+++ b/cmd/summary.go
@@ -668,8 +668,8 @@ func auditEventToProto(e events.AuditEvent) *cargowallv1.CargoWallActionEvent {
 			event.AutoAllowedType = &autoType
 		}
 	}
-	if len(e.CNAMETarget) > 0 {
-		event.CnameTarget = e.CNAMETarget
+	if len(e.CNAMEChain) > 0 {
+		event.CnameChain = e.CNAMEChain
 	}
 	return event
 }
--- a/cmd/summary.go
+++ b/cmd/summary.go
@@ -668,6 +668,9 @@ func auditEventToProto(e events.AuditEvent) *cargowallv1.CargoWallActionEvent {
 			event.AutoAllowedType = &autoType
 		}
 	}
+	if len(e.CNAMETarget) > 0 {
+		event.CnameTarget = e.CNAMETarget
+	}
 	return event
 }
 
--- a/pb/cargowall/v1/cargo_wall_action.pb.go
+++ b/pb/cargowall/v1/cargo_wall_action.pb.go
@@ -256,8 +256,12 @@ type CargoWallActionEvent struct {
 	Process         *string                        `protobuf:"bytes,8,opt,name=process,proto3,oneof" json:"process,omitempty"`
 	Category        data.CargoWallEventCategory    `protobuf:"varint,9,opt,name=category,proto3,enum=grpc.cargowall.v1.CargoWallEventCategory" json:"category,omitempty"`
 	AutoAllowedType *data.CargoWallAutoAllowedType `protobuf:"varint,10,opt,name=auto_allowed_type,json=autoAllowedType,proto3,enum=grpc.cargowall.v1.CargoWallAutoAllowedType,oneof" json:"auto_allowed_type,omitempty"`
-	unknownFields   protoimpl.UnknownFields
-	sizeCache       protoimpl.SizeCache
+	// Full CNAME chain (origin..target) when the hostname was reached via a CNAME
+	// of an allowed host; empty otherwise. The first element is the origin the
+	// user allowed, the last is the edge/target actually connected to.
+	CnameTarget   []string `protobuf:"bytes,11,rep,name=cname_target,json=cnameTarget,proto3" json:"cname_target,omitempty"`
+	unknownFields protoimpl.UnknownFields
+	sizeCache     protoimpl.SizeCache
 }
 
 func (x *CargoWallActionEvent) Reset() {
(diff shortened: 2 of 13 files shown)


mtmk (approved):
LGTM. No enforcement or security regressions; all findings are in the reporting/attribution layer and enforcement is unaffected. Approving. One item worth resolving before merge, the rest optional.

**Decide before merge (not a bug, contract cost)**
- `cname_target` (proto field 11) is singular but carries the full repeated chain. Wire/JSON name freezes on merge; renaming to `cname_chain` is free now and breaking later. Rename or consciously keep it.

**Optional, recommended**
- `pkg/dns/server.go` derived-allow path records the queried name's chain, not the final target's, so the drill-down is one hop short in the multi-hop-in-one-response case (the CDN-fronted OCSP scenario). Origin attribution stays correct.
- `pkg/events/events.go` chain[0] guard suppresses `cname_target` when the resolved hostname already equals the origin, dropping the drill-down for a subset of real chains.
- slog emits the chain as a joined string while audit/proto carry `[]string`; same field, two shapes. Passing the slice to slog aligns them.

**Optional, confirm intent**
- Block-notification dedup is now keyed on origin, so two distinct blocked edges of one origin collapse to a single alert. Intended UX or lost signal?
- Blocked-derived path on a non-inherited port now issues a redundant allow `AddIP` (already-allowed port, no security impact, just churn).
- Hostname swap runs for already-allowed connections too, which can shift the reported `auto_allowed_type`.
- Shared-CDN edge IP attributes to the most-recently-resolved origin (recency heuristic). Likely intended; a one-line comment would stop it reading as a bug later.
- `RecordCNAMEChain` is skipped on the `verdict.Matched()` branch (arguably intended).
- CNAME-learn loop takes the cache lock twice per hop (`Get` then `Merge`); `Merge` could return a pre-existed bool. Minor.

PR #74, opened 2026-06-26
#71 fix invalid block

### LINKED ISSUE #71: Shared CDN IP: all-ports rule wipes a port-specific rule's entries, blocking allowed traffic
## Summary

When a single IP is allowed by **two hostname rules with different port scopes** — one **all-ports** (`ports: []`) and one **port-specific** — the firewall can silently drop the entry's port entries and leave the IP enforced as `PortSpecific=1` **without** the ports a still-valid rule requires. The result is that legitimately-allowed traffic is **blocked at the BPF layer**, even though policy allows it.

This is hit constantly in practice because CDN/edge IPs (akamai, cloudflare, …) are shared across many hostnames, and OCSP/CRL responders are commonly configured as all-ports rules while content hosts are port-specific.

## Impact

- **Severity: high.** Legitimately-allowed egress is dropped with no policy reason. The connection retransmits and fails (build/test flakes, intermittent TLS/OCSP/CRL failures).
- Hard to diagnose: the IP shows `DNS resolution: added to firewall … action=allow` earlier in the run, yet the later connection is `Connection blocked`, and the block is reported under a **generic reverse-DNS PTR** (e.g. `a23-37-18-39.deploy.static.akamaitechnologies.com`) rather than the allowed hostname, so it looks like an unrelated host.
- The transparent-CNAME work increases exposure: more shared CDN IPs get enforced, so more of them collide with all-ports OCSP/CRL rules.

## Observed in CI

Run: `code-cargo/app` actions run `28244954165` (logs in the job summary).

```
DNS resolution: added to firewall hostname=www.microsoft.com ip=23.37.18.39 action=allow ports=[{80 tcp} {443 all}]
...
Connection blocked src=10.1.0.254:33648 dst=a23-37-18-39.deploy.static.akamaitechnologies.com dst_ip=23.37.18.39 dst_port=80 process=dotnet pid=2962 timestamp=2026-06-26 14:37:20
```

`23.37.18.39` was added for `www.microsoft.com` on `{80 tcp}` at startup, yet a TCP/80 connection to it was dropped (and retransmitted by the same src port for ~5s). A *different* www.microsoft.com IP (`23.45.137.206`, also `{80 tcp,443 all}`, added at startup) **was** allowed on :80 in the same run — the difference is that no all-ports rule ever resolved to `23.45.137.206`.

The policy contains several **all-ports** rules (`ports: []`) that are CDN-fronted and share akamai/cloudflare IPs: `ocsp.sectigo.com`, `s.symcd.com`, `ts-ocsp.ws.symantec.com`, `mpki-ocsp.digicert.com.cdn.cloudflare.net`, `blob.*.store.core.windows.net`.

## Root cause

In `pkg/firewall/firewall.go`, `addIPv4` (and the v6 twin `addIPv6`):

- Per-port entries are written additively with `UpdateAny`, and `ipPorts[ip]` tracks the cumulative union across `AddIP` calls — good for shared IPs.
- **But** the all-ports transition (`len(ports) == 0`, `firewall.go:485-499`) sets the LPM entry to `PortSpecific=0` and then **deletes every per-port entry for the IP** from `map_ports` and clears `ipPorts[ip]`:

```go
// firewall.go:485
if len(ports) == 0 {
    for _, pp := range f.ipPorts[ipStr] {
        // f.portsMap.Delete(&portKey)   // firewall.go:492 — the only runtime portsMap.Delete
    }
    delete(f.ipPorts, ipStr)
}
```

This discards ports that **other** hostname rules still require. A subsequent port-specific `AddIP` for the same IP then flips the LPM back to `PortSpecific=1` while writing only its own ports — so the previously-allowed port is gone, and the BPF egress path (`bpf/tcbpf.c`, `port_specific==1` → `map_ports` lookup miss → default deny) drops it.

This is the **only** runtime mechanism that can remove the port entry: `RemoveIP` / `removeIPFromBPFMaps` have no runtime callers, there is no periodic rebuild, and `UpdateAllowlistTC` only adds (`UpdateAny`).

### Reproduction sequence (shared IP `X`)

1. `www.microsoft.com` → `AddIP(X, allow, [80 tcp, 443])` → LPM `PortSpecific=1`, `map_ports` has `(80,tcp)`, `(443,…)`.
2. An all-ports OCSP/CRL host resolves to `X` → `AddIP(X, allow, [])` → LPM `PortSpecific=0` **and deletes** `(80,tcp)`/`(443,…)`, clears `ipPorts[X]`. *(still allowed — PS=0 is allow-all)*
3. A 443-only host (e.g. `login.microsoftonline.com`) resolves to `X` → `AddIP(X, allow, [443])` → LPM `PortSpecific=1`, `map_ports` has **only** `(443)`.
4. Connection to `X:80` → LPM hit, `PortSpecific=1`, `(80,tcp)` miss → **default deny → blocked**, despite `www.microsoft.com` allowing `X:80`.

Note a second correctness defect in the same area: step 3 **downgrades an all-ports grant** (`PortSpecific=0`) back to port-specific, so the all-ports rule's coverage is also lost.

## Proposed fix

Maintain the true union per IP so the effective BPF state reflects **all** rules that resolved to it:

- Track per IP: a count of **all-ports contributors**, and the **union of port-specific contributors' ports**.
- Effective state on every `AddIP`:
  - if all-ports count > 0 → `PortSpecific=0` (allow) — **sticky**; a later port-specific add must not flip it back to `PortSpecific=1`;
  - else → `PortSpecific=1` with the union of port-specific ports.
- Recompute on each `AddIP`; decrement/recompute on `RemoveIP`.
- Apply symmetrically to `addIPv4` and `addIPv6`.

Minimal stopgap (if a smaller change is wanted first): stop deleting at `firewall.go:492` and stop flipping `PortSpecific=0 → 1`. That alone un-blackholes the port (the retained entries make the lookup hit again), but full correctness still needs the contributor accounting above (otherwise an all-ports grant can still be silently narrowed).

## Acceptance criteria

- A shared IP allowed by `{80 tcp, 443}` (host A) and `[]` all-ports (host B) and `[443]` (host C), added in any order, allows **TCP/80, TCP/443, and any other port** afterward.
- A shared IP allowed only by `[443]` (host A) and `[80 tcp]` (host B) allows TCP/80 **and** TCP/443.
- Removing one contributor (via `RemoveIP`) leaves the remaining contributors' ports intact.
- Table tests in `pkg/firewall` covering add-order permutations for the all-ports + port-specific + narrowing sequences, for both IPv4 and IPv6.

## Notes

Discovered while working on CNAME connection-attribution reporting; this is a **separate, higher-priority** firewall correctness bug. The reporting change also helps diagnosis here (it would attribute the blocked edge IP back to `www.microsoft.com` instead of the generic akamai PTR), but does not fix the underlying drop.

matthewdevenny:
Closes #71

## Problem

When a single IP is allowed by **two hostname rules with different port scopes** — one **all-ports** (`ports: []`) and one **port-specific** — the firewall could silently drop the IP's port entries and leave it enforced as `PortSpecific=1` *without* the ports a still-valid rule requires, blocking legitimately-allowed traffic at the BPF layer. This is hit constantly in practice because CDN/edge IPs (akamai, cloudflare, …) are shared across many hostnames, and OCSP/CRL responders are commonly all-ports rules while content hosts are port-specific.

## Root cause

In `addIPv4`/`addIPv6`, the all-ports transition set `PortSpecific=0` and then **deleted every per-port entry** for the IP, and a later port-specific add **flipped `PortSpecific` back to 1** writing only its own ports — discarding ports that other rules still required (`map_ports` miss → default deny → blocked).

## Fix

Maintain a per-IP accumulator (`ipPortState`) so the effective BPF LPM entry reflects the **union of all rules** that resolved to the IP, recomputed on every write:

- **All-ports grants are sticky** — a later port-specific add can no longer narrow a `PortSpecific=0` grant; per-port entries are no longer deleted on transition.
- **Unified write paths** — `AddIP` (DNS path) and `UpdateAllowlistTC`→`addCIDR{v4,v6}` (config-load path) now share one lock-free accumulator core (`addIPv4Locked`/`addIPv6Locked`) for `/32` and `/128` host entries, so a shared IP contributed by both a static rule and a DNS resolution keeps the union instead of last-write-wins narrowing. First-touch state is seeded from any pre-existing LPM entry so the two paths can't desync.

### Conflict precedence

The BPF model can express only "all ports" (`PortSpecific=0`) or "consult `map_ports`" (`PortSpecific=1`) — never "allow all except port N". When contributions to a shared IP conflict:

- **all-ports deny beats all-ports allow** — a host the policy denies outright is not re-opened by an unrelated hostname sharing the address.
- an **all-ports grant beats a port-specific contribution of the opposite action** — this preserves #71's "don't block legitimately-allowed traffic" guarantee. The unenforceable per-port entry is write-skipped (no wasted syscalls, no inert `map_ports` slots, no spurious "changed") and the dropped intent is logged with a `Warn` so it isn't silent.

`events.go`'s late-allow audit comment is corrected to note it predicts from a single hostname's verdict and can diverge from the BPF verdict on such shared-IP edges (audit-only; enforcement unaffected).

## Acceptance criteria (all covered by tests)

- A shared IP allowed by `{80 tcp, 443}` + `[]` all-ports + `[443]`, added in **any order**, allows TCP/80, TCP/443, and any other port — IPv4 and IPv6.
- A shared IP allowed only by `[443]` + `[80 tcp]` allows TCP/80 **and** TCP/443 (and stays port-specific — other ports denied).
- `RemoveIP` cleans the full union and leaves unrelated IPs intact.
- All-ports deny wins over all-ports allow regardless of order; deny-all blocks even under a default-allow policy.
- The config-load and runtime paths agree for a shared all-ports + port-specific IP in either order.

## Testing

Table-driven tests cover all add-order permutations for both v4 and v6, a faithful in-Go mirror of the `tcbpf.c` decision path (`bpfDecideV4/V6`) to assert the real allow/deny outcome, plus the precedence, inert-port, and unification cases. `go build` / `go vet` / `staticcheck` / `gofumpt` are clean for `GOOS=linux`; the firewall unit tests run on Linux/CI (eBPF build tags).

### DIFF
--- a/pkg/events/events.go
+++ b/pkg/events/events.go
@@ -366,19 +366,27 @@ func processEvent(raw []byte, configMgr *config.Manager, notificationTracker *No
 						logger.Debug("Late-resolved IP already in firewall",
 							"ip", dstIP, "hostname", hostname, "ports", verdict.AllowPorts)
 					}
-					// Sound regardless of `changed`: FirewallImpl.addIPv4 /
-					// addIPv6 reconcile per-port entries before the LPM no-op
-					// check, so on err==nil the current rule's `ports` are
-					// guaranteed to be in map_ports — even when the IP was
-					// already in the LPM from a different rule with disjoint
-					// ports. Checking the inputs alone therefore answers
-					// "will the retry succeed?" correctly.
+					// Best-effort prediction of "will the retry succeed?" from
+					// this rule's own ports: FirewallImpl reconciles per-port
+					// entries before the LPM no-op check, so on err==nil the
+					// current rule's `ports` are in map_ports even when the IP
+					// was already in the LPM from a different rule with disjoint
+					// ports.
 					//
 					// For a mixed verdict (e.g. `*.foo: deny 80` + `bar:
 					// allow all` on `bar.foo`), AllowPorts may be empty (all
 					// ports) while DenyPorts covers the event's port. The
 					// retry on that port will still be blocked by the deny
 					// side's per-port BPF entry, so it is NOT late-allowed.
+					//
+					// Caveat: this looks only at THIS hostname's verdict, not at
+					// other rules that resolved to the same shared IP. If that IP
+					// already carries a conflicting all-ports grant (e.g. a
+					// different all-ports-deny host shares it), the firewall's
+					// PortSpecific=0 stickiness makes this rule's per-port entry
+					// inert, so the audited late-allow/blocked label can diverge
+					// from the actual BPF verdict for that edge. Enforcement is
+					// unaffected — this only governs the audit/notification.
 					allowMatches := dstPortAllowedByRule(event.DstPort, event.IpProto, verdict.AllowPorts)
 					denyMatches := verdict.HasDeny() &&
 						dstPortAllowedByRule(event.DstPort, event.IpProto, verdict.DenyPorts)
--- a/pkg/firewall/firewall.go
+++ b/pkg/firewall/firewall.go
@@ -48,7 +48,13 @@ type Firewall interface {
 	// identical state returns (false, nil). Per-port entries accumulate across
 	// calls (set semantics): the same IP added under two hostname rules with
 	// disjoint port sets ends up with both port sets in the map, and RemoveIP
-	// will clean up all of them.
+	// will clean up all of them. An all-ports add (empty ports) marks the IP as
+	// all-ports; that PortSpecific=0 state is sticky, so a later port-specific
+	// add for the same IP cannot silently narrow it back to specific ports.
+	// Conflict precedence on a shared IP: all-ports deny beats all-ports allow,
+	// and any all-ports grant beats a port-specific contribution of the opposite
+	// action (the port entry is written-skipped and the dropped intent logged) —
+	// see ipPortState.
 	AddIP(ip net.IP, action config.Action, ports []config.Port) (bool, error)
 
 	// RemoveIP removes a single IP from the BPF maps
@@ -73,8 +79,10 @@ type FirewallImpl struct {
 	logger           *slog.Logger
 	mu               sync.RWMutex
 
-	// Track which ports are associated with each IP for proper cleanup
-	ipPorts map[string][]portProto // IP string -> port+protocol pairs
+	// Per-IP accumulated firewall state, keyed by IP string. Tracks the union of
+	// port-specific entries (for cleanup) plus the sticky all-ports grant, so a
+	// shared IP's effective LPM entry reflects every rule that resolved to it.
+	ipPorts map[string]*ipPortState
 }
 
 // portProto carries a port number and its IANA protocol number for BPF map keys.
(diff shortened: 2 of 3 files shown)


mtmk (approved):
LGTM

PR #78, opened 2026-07-07
#77 fix cname dedupe

### LINKED ISSUE #77: summary: cname_chain dropped by deduplicateStepEvents (chainless duplicate to the same origin wins)
## Summary

`cargowall summary` drops the `cname_chain` on connection events before pushing them to the API. `deduplicateStepEvents` collapses events on a key that does **not** include `CNAMEChain`, and it keeps the **first-seen** event. Because the transparent-CNAME feature attributes a CNAME-derived connection to the **origin hostname** (the same hostname a direct connection to that origin reports), a chain-bearing event and a chainless direct connection to the same origin share a dedup key. Whichever sorts earlier wins — and when the chainless one is first, the CNAME chain is silently lost.

Net effect: any run where traffic reaches both an allowed origin **and** a CNAME target of that origin loses the `cname_chain` drill-down for that destination. The firewall records it correctly; it's discarded in the summary/push stage.

## Affected version

Observed on `v1.3.2`; still present on `main` (`4b7ca6d43a63309e7e04360286cee9b18bc222cf`).

## Root cause

`deduplicateStepEvents` runs before the API push:

- [`cmd/summary.go:138`](https://github.com/code-cargo/cargowall/blob/4b7ca6d43a63309e7e04360286cee9b18bc222cf/cmd/summary.go#L138) — `deduplicateStepEvents(stepEvents)` then `pushToApi(...)`
- The dedup key omits `CNAMEChain`: [`cmd/summary.go:259-265`](https://github.com/code-cargo/cargowall/blob/4b7ca6d43a63309e7e04360286cee9b18bc222cf/cmd/summary.go#L259-L265)

```go
type dedupKey struct {
	process   string
	dest      string
	port      uint16
	protocol  string
	eventType events.AuditEventType
}
```

- First-seen wins: [`cmd/summary.go:267-282`](https://github.com/code-cargo/cargowall/blob/4b7ca6d43a63309e7e04360286cee9b18bc222cf/cmd/summary.go#L267-L282)

```go
key := dedupKey{process: event.Process, dest: dest, port: event.DstPort, protocol: event.Protocol, eventType: event.EventType}
if _, exists := seen[key]; !exists {
	seen[key] = struct{}{}
	deduped = append(deduped, event)
}
```

A CNAME-derived connection is reported under the origin hostname (`chain[0]`), so `dest` for the chain event equals the origin hostname — identical to the `dest` of a plain connection to that origin. Same `process`/`port`/`protocol`/`eventType` ⇒ identical `dedupKey`. Events are processed in timestamp order, so a direct hit on the origin that precedes the CNAME-target hit is kept, and the chain-bearing event is dropped as a "duplicate." The retained event's `CNAMEChain` is empty, so `auditEventToProto` ([`cmd/summary.go:671-672`](https://github.com/code-cargo/cargowall/blob/4b7ca6d43a63309e7e04360286cee9b18bc222cf/cmd/summary.go#L671-L672)) never sets `cname_chain`, and the pushed/stored event has no chain.

More generally, the dedup discards the distinguishing attributes (`cname_chain`, and potentially `matched_rule` / `auto_allowed_type`) of every later duplicate — whichever event sorts first defines the reported attributes for that `(process, dest, port, protocol, type)` group. `cname_chain` is the visible casualty.

## Reproduction

In a GitHub Actions job with `cargowall-action`:

1. Allow an origin whose DNS resolves cross-domain via CNAME to a target that is **not** independently allowed (e.g. allow `docker.io`; `auth.docker.io` → CNAME `auth.docker.io.cdn.cloudflare.net`, with `cloudflare.net` not allowed).
2. Make a direct connection to the origin (`curl https://auth.docker.io/...`).
3. Then connect to the CNAME target by name (`curl https://auth.docker.io.cdn.cloudflare.net/`), which the firewall derives-allows and attributes back to `auth.docker.io` with a chain.

Both connections are `process=curl`, `dest=auth.docker.io`, `port=443`, `tcp`, `ConnectionAllowed` → same dedup key. The step-1 event (no chain) is kept; the step-3 event (chain) is dropped.

### Observed evidence

Firewall log (event is present, with the chain):

```
Connection allowed src=... dst=auth.docker.io dst_ip=104.18.43.178 dst_port=443 process=curl \
  cname_chain=[auth.docker.io auth.docker.io.cdn.cloudflare.net]
```

But the audit push for that run stored the `auth.docker.io` event with an **empty** `cname_chain` (verified against the CodeCargo API response for the run). Resolving the origin by DNS only (no TCP connect) so the CNAME-target connection is the *sole* `auth.docker.io:443` event makes the chain survive — confirming dedup as the cause.

## Suggested fix

**Preferred — keep one row per destination, but never lose a chain.** On a dedup collision, upgrade the retained event when a later duplicate carries a richer `CNAMEChain` (and, ideally, other distinguishing fields it's missing):

```go
func deduplicateStepEvents(stepEvents []StepEvents) {
	for i, se := range stepEvents {
		seen := make(map[dedupKey]int) // key -> index into deduped
		var deduped []events.AuditEvent
		for _, event := range se.Events {
			dest := event.DstHostname
			if dest == "" {
				dest = event.DstIP
			}
			key := dedupKey{process: event.Process, dest: dest, port: event.DstPort, protocol: event.Protocol, eventType: event.EventType}
			if idx, exists := seen[key]; exists {
				// Prefer the representative that carries a CNAME chain.
				if len(deduped[idx].CNAMEChain) == 0 && len(event.CNAMEChain) > 0 {
					deduped[idx].CNAMEChain = event.CNAMEChain
				}
				continue
			}
			seen[key] = len(deduped)
			deduped = append(deduped, event)
		}
		stepEvents[i].Events = deduped
	}
}
```

**Alternative — treat them as distinct.** Add the chain to the key (e.g. `chain: strings.Join(event.CNAMEChain, ">")`) so a via-CNAME connection isn't collapsed with a direct one. Downside: emits a separate, redundant chainless row for the same destination when both occur.

The merge approach is likely what dashboards want (one row per destination, chain preserved).

matthewdevenny:
Closes #77

## Why

`cargowall summary` dedupes connection events per step before the SaaS push, on a key of `(process, dest, port, protocol, eventType)` — and keeps the **first-seen** event wholesale. Transparent CNAME support (#63/#68/#73) attributes a CNAME-derived connection to its **origin hostname** (`chain[0]`), the same `dest` a direct connection to that origin reports. So a chain-bearing event and a chainless direct hit to the same origin collide on the dedup key, and whenever the chainless one sorts first, the chain-bearing event is discarded — the pushed/stored event has an empty `cname_chain` even though the firewall logged the chain correctly (see #77 for the observed `auth.docker.io` evidence). The same wholesale discard applies to `auto_allowed_type`, which is per-IP (CIDR pass of `GetAutoAllowedType`) while the dedup key ignores `DstIP` when a hostname is set — round-robin DNS can produce same-key duplicates where only a later one carries the type.

## What

- **`cmd/summary.go` — `deduplicateStepEvents`:** the `seen` map now stores the index of the retained representative (`map[dedupKey]int`). On a collision, instead of discarding the duplicate outright, backfill the distinguishing fields the representative is missing:
  - `CNAMEChain` — filled when the representative has none and the duplicate carries one;
  - `AutoAllowedType` — same fill-if-empty rule.

  Fill-if-empty only: the first-seen event stays the representative and an existing value is never overwritten. One row per destination is preserved (the merge approach preferred in #77, not the add-chain-to-key alternative, which would emit a redundant chainless row). No signature or caller changes — `auditEventToProto` picks up the now-populated fields via its existing non-empty guards.

  `MatchedRule` deliberately gets **no** backfill: only `connection_late_allowed` events carry it, and they always do (`HasAllow()` ⇒ non-empty rule), so with `eventType` in the dedup key a same-key group is uniformly populated or uniformly empty — the backfill branch could never fire. Documented in a comment.

- **`cmd/summary_test.go`:** three new tests alongside the existing dedup group —
  - `ChainBackfilledFromLaterDuplicate` — the #77 repro shape: chainless direct hit first, chain-bearing CNAME-derived duplicate second → one row, chain preserved, first-seen timestamp retained;
  - `ChainNotOverwritten` — a later chainless duplicate doesn't clobber an existing chain;
  - `AutoAllowedTypeBackfilled` — per-IP round-robin shape: type backfilled from the first duplicate that carries one, a later different type doesn't overwrite.

## Semantics note

For a merged row with mixed auto/non-auto duplicates, the auto-allow attribution now deterministically wins (previously the label was ordering-arbitrary, first-seen verbatim). This means `totalAutoAllowed` / `AutoAllowedConnections` count a destination as auto-allowed if **any** connection to it was — intended, since at least one connection genuinely took the auto-allow path, and the alternative silently hides that attribution.

## Testing

- All 10 `TestSummary_DeduplicateStepEvents_*` tests (7 existing + 3 new) and the full `cmd` package pass on linux (`golang:1.25` container; the dev host is darwin and these files are `//go:build linux`).
- `GOOS=linux go build ./...`, `go vet`, `staticcheck`, and `gofumpt -l` are clean.

### DIFF
--- a/cmd/summary.go
+++ b/cmd/summary.go
@@ -264,31 +264,34 @@ type dedupKey struct {
 	eventType events.AuditEventType
 }
 
+// backfillDistinguishingFields keeps one row per destination while preserving
+// distinguishing fields carried only by a later duplicate (e.g. a CNAME chain
+// or auto-allowed type — see issue #77). Only empty fields on the retained
+// representative are filled; a value already present is authoritative.
+// MatchedRule needs no backfill: only connection_late_allowed events carry it,
+// and they always do, so with eventType in the dedup key a same-key group is
+// uniformly populated or uniformly empty.
+func backfillDistinguishingFields(rep, dup *events.AuditEvent) {
+	if len(rep.CNAMEChain) == 0 && len(dup.CNAMEChain) > 0 {
+		rep.CNAMEChain = dup.CNAMEChain
+	}
+	if rep.AutoAllowedType == "" && dup.AutoAllowedType != "" {
+		rep.AutoAllowedType = dup.AutoAllowedType
+	}
+}
+
 func deduplicateStepEvents(stepEvents []StepEvents) {
-	for i, se := range stepEvents {
+	for i := range stepEvents {
 		seen := make(map[dedupKey]int) // key -> index into deduped
 		var deduped []events.AuditEvent
-		for _, event := range se.Events {
+		for _, event := range stepEvents[i].Events {
 			dest := event.DstHostname
 			if dest == "" {
 				dest = event.DstIP
 			}
 			key := dedupKey{process: event.Process, dest: dest, port: event.DstPort, protocol: event.Protocol, eventType: event.EventType}
 			if idx, exists := seen[key]; exists {
-				// Keep one row per destination, but backfill distinguishing
-				// fields the retained representative is missing (e.g. a CNAME
-				// chain or auto-allowed type carried only by a later duplicate
-				// — see issue #77). Only fill empty fields; a value already on
-				// the representative is authoritative. MatchedRule needs no
-				// backfill: only connection_late_allowed events carry it, and
-				// they always do, so with eventType in the key a same-key group
-				// is uniformly populated or uniformly empty.
-				if len(deduped[idx].CNAMEChain) == 0 && len(event.CNAMEChain) > 0 {
-					deduped[idx].CNAMEChain = event.CNAMEChain
-				}
-				if deduped[idx].AutoAllowedType == "" && event.AutoAllowedType != "" {
-					deduped[idx].AutoAllowedType = event.AutoAllowedType
-				}
+				backfillDistinguishingFields(&deduped[idx], &event)
 				continue
 			}
 			seen[key] = len(deduped)
--- a/cmd/summary.go
+++ b/cmd/summary.go
@@ -266,7 +266,7 @@ type dedupKey struct {
 
 func deduplicateStepEvents(stepEvents []StepEvents) {
 	for i, se := range stepEvents {
-		seen := make(map[dedupKey]struct{})
+		seen := make(map[dedupKey]int) // key -> index into deduped
 		var deduped []events.AuditEvent
 		for _, event := range se.Events {
 			dest := event.DstHostname
@@ -274,10 +274,25 @@ func deduplicateStepEvents(stepEvents []StepEvents) {
 				dest = event.DstIP
 			}
 			key := dedupKey{process: event.Process, dest: dest, port: event.DstPort, protocol: event.Protocol, eventType: event.EventType}
-			if _, exists := seen[key]; !exists {
-				seen[key] = struct{}{}
-				deduped = append(deduped, event)
+			if idx, exists := seen[key]; exists {
+				// Keep one row per destination, but backfill distinguishing
+				// fields the retained representative is missing (e.g. a CNAME
+				// chain or auto-allowed type carried only by a later duplicate
+				// — see issue #77). Only fill empty fields; a value already on
+				// the representative is authoritative. MatchedRule needs no
+				// backfill: only connection_late_allowed events carry it, and
+				// they always do, so with eventType in the key a same-key group
+				// is uniformly populated or uniformly empty.
+				if len(deduped[idx].CNAMEChain) == 0 && len(event.CNAMEChain) > 0 {
+					deduped[idx].CNAMEChain = event.CNAMEChain
+				}
+				if deduped[idx].AutoAllowedType == "" && event.AutoAllowedType != "" {
+					deduped[idx].AutoAllowedType = event.AutoAllowedType
+				}
+				continue
 			}
+			seen[key] = len(deduped)
+			deduped = append(deduped, event)
 		}
 		stepEvents[i].Events = deduped
 	}
--- a/cmd/summary_test.go
+++ b/cmd/summary_test.go
@@ -175,6 +175,27 @@ func TestSummary_DeduplicateStepEvents_ChainBackfilledFromLaterDuplicate(t *test
 	assert.Equal(t, ts, stepEvents[0].Events[0].Timestamp, "representative is still the first-seen event")
 }
 
+// The backfill is unconditional on EventType — the Log*Blocked paths also
+// emit chains (e.g. a derived connection blocked on a non-inherited port is
+// attributed to the origin with its chain), so a chain carried by a later
+// blocked duplicate must be preserved too.
+func TestSummary_DeduplicateStepEvents_ChainBackfilledOnBlockedEvent(t *testing.T) {
+	ts := time.Date(2025, 1, 1, 0, 0, 0, 0, time.UTC)
+	chain := []string{"auth.docker.io", "auth.docker.io.cdn.cloudflare.net"}
+	stepEvents := []StepEvents{
+		{
+			Step: GitHubStep{Name: "build"},
+			Events: []events.AuditEvent{
+				{Timestamp: ts, EventType: events.EventConnectionBlocked, DstHostname: "auth.docker.io", DstIP: "1.1.1.1", DstPort: 25, Protocol: "TCP", Process: "curl"},
+				{Timestamp: ts.Add(time.Second), EventType: events.EventConnectionBlocked, DstHostname: "auth.docker.io", DstIP: "2.2.2.2", DstPort: 25, Protocol: "TCP", Process: "curl", CNAMEChain: chain},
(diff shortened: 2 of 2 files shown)


mtmk (approved):
LGTM

cmd/summary.go
```diff
@@ -266,18 +266,33 @@ type dedupKey struct {
...
 		for _, event := range se.Events {
 			dest := event.DstHostname
 			if dest == "" {
 				dest = event.DstIP
 			}
 			key := dedupKey{process: event.Process, dest: dest, port: event.DstPort, protocol: event.Protocol, eventType: event.EventType}
-			if _, exists := seen[key]; !exists {
-				seen[key] = struct{}{}
-				deduped = append(deduped, event)
+			if idx, exists := seen[key]; exists {
```
matthewdevenny, line 293: **[suggestion]** The backfill logic and its rationale are implemented via a manual if-block + comment for exactly two fields. If `AuditEvent` gains additional per-observation fields that can legitimately differ across same-key duplicates (or if `MatchedRule` semantics ever change for late-allowed events), the list and comment will need synchronized updates to avoid silent data loss in summaries and API pushes.

**Suggestion:** Consider extracting a small helper (e.g. `backfillDistinguishingFields(rep, dup *events.AuditEvent)`) or a list of backfillable fields so the policy is expressed in one place. Keep the "only fill if empty; first is authoritative" rule.

cmd/summary.go
```diff
@@ -266,18 +266,33 @@ type dedupKey struct {
 
 func deduplicateStepEvents(stepEvents []StepEvents) {
 	for i, se := range stepEvents {
-		seen := make(map[dedupKey]struct{})
+		seen := make(map[dedupKey]int) // key -> index into deduped
```
matthewdevenny, line 285: **[nit]** The in-place mutation of `stepEvents[i].Events` (and of struct values inside the `deduped` slice) is correct and matches prior behavior, but the `for i, se := range` + later index write pattern is slightly indirect. No clones or extra allocations beyond the necessary struct copies on append.

**Suggestion:** A minor readability alternative is `for i := range stepEvents { ... stepEvents[i].Events = deduped }` (still using the index write). Not required for correctness.

cmd/summary_test.go
```diff
@@ -151,6 +151,74 @@ func TestSummary_DeduplicateStepEvents_DifferentProtocolsKept(t *testing.T) {
...
+				{Timestamp: ts.Add(time.Second), EventType: events.EventConnectionAllowed, DstHostname: "auth.docker.io", DstIP: "1.1.1.1", DstPort: 443, Protocol: "TCP", Process: "curl"},
+			},
+		},
+	}
+	deduplicateStepEvents(stepEvents)
+	require.Len(t, stepEvents[0].Events, 1)
+	assert.Equal(t, chain, stepEvents[0].Events[0].CNAMEChain, "retained chain must be preserved")
+}
+
+// auto_allowed_type is per-IP (CIDR pass of GetAutoAllowedType) while the
```
matthewdevenny, line 218: **[suggestion]** The new tests (and the pre-existing protocol test) document why `DstIP` is ignored in the dedup key and why backfill is required. However, no test case exercises a backfill scenario for `CNAMEChain` on a *blocked* or *protocol-blocked* event (even though the backfill code path is unconditional on `EventType` and `Log*Blocked` paths also emit `cnameChain`).

**Suggestion:** Add one minimal test variant (or extend an existing) using `EventConnectionBlocked` (or `EventProtocolBlocked`) carrying a chain on the later duplicate. This would increase confidence that the enrichment applies to the full set of event types that carry chains.

cmd/summary.go
```diff
@@ -266,18 +266,33 @@ type dedupKey struct {
...
+				// Keep one row per destination, but backfill distinguishing
+				// fields the retained representative is missing (e.g. a CNAME
+				// chain or auto-allowed type carried only by a later duplicate
+				// — see issue #77). Only fill empty fields; a value already on
+				// the representative is authoritative. MatchedRule needs no
+				// backfill: only connection_late_allowed events carry it, and
+				// they always do, so with eventType in the key a same-key group
+				// is uniformly populated or uniformly empty.
+				if len(deduped[idx].CNAMEChain) == 0 && len(event.CNAMEChain) > 0 {
+					deduped[idx].CNAMEChain = event.CNAMEChain
```
matthewdevenny: **[nit]** Direct slice-header assignment `deduped[idx].CNAMEChain = event.CNAMEChain` (and the struct append) shares backing arrays with the source event. Current usage never mutates these slices post-construction, so this is safe and avoids an unnecessary clone.

**Suggestion:** Leave as-is (no clone); if a future change ever appends to `CNAMEChain` after creation, add a comment here or perform a copy at that time. No action needed for this PR.

PR #84, opened 2026-07-16
#83 reconcile late-allowed blocks and pre-resolve CNAME-only allowed responses

### LINKED ISSUE #83: Late-allowed IPs shipped to SaaS as blocks when A records bypass the DNS proxy
## Symptom

In run https://github.com/code-cargo/app/actions/runs/29470151967 (Checks job), the CargoWall Check failed on 3 IPs (`20.209.112.225`, `20.209.178.193`, `20.209.113.193`) belonging to `productionresultssa5.blob.core.windows.net` — a hostname that matches the auto-added `blob.core.windows.net` allow rule and whose CNAME chain had already been learned. The connections were eventually allowed, but ~40 blocked SYN retries were shipped to the SaaS as denies.

## Timeline (from the run log)

1. `03:57:00.043` — proxy processes a **CNAME-only** response for `productionresultssa5.blob.core.windows.net` (likely the AAAA query; Azure blob has no AAAA). Chain learned, but no A/AAAA records → no IPs added to the BPF allow map.
2. The client already had the three A-record IPs (connected 6 ms later) — the A answer never traversed the proxy, most plausibly served from systemd-resolved's local cache. `prePopulateDNSCache` couldn't help: it only pre-resolves rule hostnames (the `blob.core.windows.net` apex, which doesn't resolve to per-storage-account IPs).
3. `03:57:00–03:57:23` — every SYN retry blocked. In `handleConnectionEvent`, every attribution path misses (`LookupHostnameByIP`, PTR, `ForwardMatchIP`, `LookupCNAMEChain` — CNAME knowledge is name-keyed; per-IP records are only written when A records flow through the proxy). With `hostname == ""` the late-allow path never runs → plain `connection_blocked` events with raw IPs.
4. `03:57:35.789` — client re-resolves after cache expiry, A response finally passes through the proxy, IPs added, connection allowed. The earlier blocks are already in the append-only audit log; nothing reconciles them. `summary` maps every `connection_blocked` → DENY → check fails.

## Root causes

- **Reporting**: blocks that are later allowed (same IP/port/proto added to the firewall via DNS resolution) are still shipped to the SaaS as denies. The existing `connection_late_allowed` mechanism only fires when the IP is attributable to an allowed hostname *at block time*.
- **Enforcement window**: an allowed response that carries only a CNAME chain (no address records) leaves the resolved IPs unenforced until a later query happens to traverse the proxy.

## Fix

1. **Late-allow reconciliation**: keep a short-TTL buffer of recently blocked `(dst_ip, dst_port, proto)` in the daemon. When the DNS path adds an allow IP to the BPF maps, emit `connection_late_allowed` audit events for matching recent blocks (now with hostname/rule attribution). In `summary`, reclassify/drop earlier `connection_blocked` events that share `dst_ip+port+proto` with a later `connection_late_allowed` event so they are not pushed as denies.
2. **CNAME-only pre-resolution**: when a rule-allowed or derived-allowed response is CNAME-only (chain learned, zero IPs extracted), asynchronously resolve A/AAAA for the terminal target through the proxy's upstream and apply the results, closing the window before the client's first retry.

matthewdevenny:
Closes #83

## Problem

In [this run](https://github.com/code-cargo/app/actions/runs/29470151967), the CargoWall Check failed on 3 IPs of `productionresultssa5.blob.core.windows.net` — a hostname covered by the auto-added `blob.core.windows.net` allow rule whose CNAME chain had already been learned. The client got the A records from a resolution path that never traversed the DNS proxy (warm systemd-resolved cache; only the CNAME-only AAAA answer came upstream), so:

1. The IPs were never added to the BPF allow map — ~40 SYN retries were blocked over 23s.
2. The blocked events couldn't be attributed to a hostname (no A record seen, no PTR, CNAME knowledge is name-keyed), so the existing in-band late-allow path never fired and they were audited as plain `connection_blocked` with raw IPs.
3. When the client re-resolved 35s later, the A response finally flowed through the proxy and the IPs were allowed — but the blocks were already in the append-only audit log and shipped to the SaaS as denies, failing the check.

Auditing the same startup window also surfaced three holes in how pre-existing connections are handled when cargowall attaches (fixes 3–5 below): the datapath is stateless and destination-based, so an established connection survives attach only if its remote IP is seeded into the allow maps first — and the seeding logic had gaps.

## Fix 1 — late-allow reconciliation

- **`pkg/events`**: new `RecentBlocks` buffer, subscribed to the audit stream as an `EventSink`. Keeps the latest blocked TCP/UDP attempt per `(dst_ip, dst_port, protocol)` (5 min TTL, 4096-entry cap, fail-open on overflow).
- **`pkg/dns`**: whenever the enforcement path opens the firewall for an IP (rule match, derived CNAME allow, or the rules-arrived-late backfill in `ApplyRulesToTrackedHostnames`), it drains matching buffered blocks and emits `connection_late_allowed` audit events for them — dated at the original blocked attempt, attributed to the allowed hostname/rule, with the CNAME chain as drill-down. Mixed verdicts are respected (`allowMatches && !denyMatches`, same rule as the in-band path) — including on the derived CNAME path, which passes the origin's deny side into reconciliation. `applyVerdictSide` reports whether the allow was actually written, so a failed BPF write can't mislabel a block, and the default-action short-circuit (`defaultAction=allow`) reports false too: nothing was opened, and a shared-IP deny entry may still block the peer.
- **`cmd/summary`**: `reconcileLateAllowedBlocks` drops `connection_blocked` events superseded by a late-allowed event for the same tuple at the same time or later, so they aren't counted or pushed as denies. Blocks *after* the last late-allow for a tuple are kept (genuinely re-blocked).

## Fix 2 — CNAME-only pre-resolution

- **`pkg/dns`**: an allowed response that carries only a CNAME chain (the incident shape) now triggers a background A/AAAA resolution of the terminal target through the upstream, run through the normal enforcement path — the firewall opens within milliseconds instead of waiting for the client's cache to expire. Deduped per target (30s TTL) and depth-bounded (one recursion for chains split across round-trips), so a malicious never-terminating chain can't loop.

With both fixes, the incident plays out as: CNAME-only AAAA response → chain learned → terminal target pre-resolved and IPs allowed ~instantly; any block that raced the resolution is re-reported as late-allowed and reconciled out of the deny counts.

## Fix 3 — `--allow-existing-connections` decoupled from `--prepopulate-dns-cache` + attach-race closed

- **`cmd/start.go`**: the `/proc` connection scan only ran inside the `--prepopulate-dns-cache` block, so `--allow-existing-connections` alone was silently a no-op. The scan (which needs no DNS) now runs when either flag is set; only the reverse-DNS enrichment stays gated on prepopulation.
- The scan also ran *before* up to ~35s of sequential DNS prepopulation lookups, so connections established in that window were killed mid-stream the moment the TC program attached. A cheap re-scan immediately before `tc.AttachEgress` now gates just the delta — excluded at tuple granularity, so a new port on an already-seen peer still surfaces — reusing the same idempotent, audited gate.

## Fix 4 — broader, tuple-scoped existing-connection scan

- **`cmd/start.go`**: `scanProcTCP` → `scanProcNet` with a state-set parameter. TCP now keeps ESTABLISHED plus in-flight handshakes (SYN_SENT/SYN_RECV), and connected UDP sockets (`/proc/net/udp{,6}`, QUIC/VPN-style long-lived flows) are scanned too. Zero/unspecified remotes are skipped alongside the existing loopback/link-local filters.
- The scan returns `(ip, port, protocol)` tuples, and gating uses them: rule-matched peers keep their rule's ports as before, but unresolvable/unmatched peers are now allowed **only on their observed remote ports** instead of the previous IP-wide all-ports carveout. A pre-existing socket only needs its own destination tuple to stay alive; a *new* connection to a different port on the same peer correctly falls to default policy.

## Fix 5 — observable mid-stream drops

Previously the BPF blocked path emitted events only for TCP SYN and UDP, so an established connection killed mid-stream (e.g. a pre-existing socket whose destination was never seeded) dropped every packet silently — invisible to the audit log and unrecoverable by the Fix 1 reconciler.

- **`bpf/tcbpf.c`**: the event struct's padding byte at offset 3 becomes a `flags` byte (wire layout unchanged); non-SYN TCP drops now emit events flagged `EVENT_FLAG_MIDSTREAM` on both IPv4 and IPv6 paths. Two LRU maps rate-limit emission to one event per destination tuple per 10s, so retransmit storms (and audit-mode data segments) can't flood the ring buffer.
- **`pkg/events` / `pkg/otlp`**: these surface as `connection_blocked` with `mid_stream: true` (and a `cargowall.mid_stream` OTLP attribute). Keeping the event type unchanged means the `RecentBlocks` reconciler and the in-band late-allow path recover killed connections for free — when the destination maps to an allowed hostname, the firewall re-opens and retransmits survive.

## Testing

- Unit tests for `RecentBlocks` (filtering, latest-attempt-wins, port/protocol matching, deny exclusion, TTL, take-once semantics, audit-sink integration, mid-stream recovery) and `reconcileLateAllowedBlocks` (supersession windows, unrelated tuples untouched).
- End-to-end DNS tests: incident replay (block → resolution → reconciled late-allow with original timestamp, matched rule, attribution) and CNAME-only pre-resolve against a local test upstream, including derived-path origin attribution and the dedup guard.
- Table tests for `scanProcNet` (states kept/skipped per protocol, zero/loopback/link-local remotes, malformed lines, IPv6 group decoding) and the delta-scan exclude filter.
- Startup-wiring regression tests via injectable `/proc` sources: the scan runs for either flag alone (the original `--allow-existing-connections` no-op bug), gated tuples are recorded for the delta re-scan, the pre-attach re-scan gates only tuples that appeared after the initial scan (including a new port on an already-gated peer), unmatched peers get observed-ports-only allows, flag-off paths never touch the firewall, and missing optional sources (IPv6 disabled) don't fail the scan. The OTLP mapper's `cargowall.mid_stream` attribute is asserted present on mid-stream blocks and absent otherwise.
- Reconciliation-labeling tests for the review findings: under `defaultAction=allow` a matching allow verdict emits no late-allowed events and leaves the buffered block untouched; a derived CNAME allow with a mixed-verdict origin late-allows only allow-covered blocks and silently drops origin-denied-port blocks.
- BPF tests: mid-stream drops emit flagged events, repeats on the same tuple are rate-limited, distinct tuples emit, SYN blocks stay unflagged, and audit mode passes the packet while still emitting (rate-limited).
- Full suite passes on linux including the sudo BPF/tc tests; `gofumpt`, `goimports-reviser`, `staticcheck`, `go vet` clean.


mtmk (approved):
LGTM.

Optional consider: the CNAME pre-resolve goroutine has no shutdown context or `recover()` (fire-and-forget on `s.client.Exchange`) - a hang leaks it and a panic on the resolved response takes down the daemon. Not a blocker.

PR #90, opened 2026-07-21
#89 flush cache at eBPF attach

### LINKED ISSUE #89: systemd-resolved stub cache serves pre-attach answers invisibly, causing unattributed bare-IP blocks (flush at attach)
On a GitHub Actions runner (v1.3.4, Azure, enforce mode), the runner's results uploader was blocked connecting to bare Azure Storage IPs with no hostname attribution and no preceding DNS query:

```
16:21:43.628 Connection blocked src=10.1.0.191:38698 dst=20.150.82.228 dst_ip=20.150.82.228 dst_port=443 process=MainThread pid=2506
16:21:43.649 Connection blocked src=10.1.0.191:56492 dst=20.150.88.228 dst_ip=20.150.88.228 dst_port=443 process=MainThread pid=2506
```

`20.150.0.0/15` is Azure Storage; pid 2506 is the runner's results-upload thread (same process talking to `results-receiver.actions.githubusercontent.com`). These are `productionresultssa*.blob.core.windows.net` frontends — a hostname the auto-added `blob.core.windows.net` infra rule would allow, had the resolution traversed the proxy. The job survived only because an already-open connection to `52.239.172.36` had been caught by the pre-existing-connection allow; the blocks shipped to the SaaS as apparent policy violations.

## Mechanism

The client never queried through the proxy because **systemd-resolved's stub kept serving its warm pre-attach cache**. The DNS redirect intercepts resolved's *upstream* queries — but a stub cache hit (127.0.0.53) never goes upstream, so the proxy never sees the name, no hostname↔IP mapping is learned, and the connection arrives as a bare IP → deny-by-default.

pid 2506 is python, which has no in-process DNS cache: its `getaddrinfo` went to the stub every time and was answered invisibly from resolved's cache, populated before the cargowall step ran (the runner boots first).

Neither existing startup mitigation can reach this:

- **Pre-existing connection scan** covers sockets already open at attach — not future connections to cached IPs.
- **System DNS cache snapshot** (`prePopulateDNSCache` phase 1, `cmd/start.go`) queries the stub for *tracked hostnames only*. A name matched by a wildcard/suffix rule (`blob.*.store.core.windows.net`, `blob.core.windows.net`) is unknowable in advance — the concrete storage-account name only exists once someone resolves it — so the snapshot structurally cannot cover this class.

The current behavior is a TTL lottery: an earlier run (07-17) resolved `productionresultssa13.blob.core.windows.net` through the proxy and was allowed, because resolved's cache happened to be cold for that name; this run's cache was warm, so it wasn't.

## Fix

Flush systemd-resolved's cache once the interception is in place, so every subsequent lookup from any process misses the stub, goes upstream, and traverses the proxy — where suffix rules match, IPs are firewall-allowed, and attribution is kept:

```
install DNS redirect → snapshot tracked hostnames (as today) → resolvectl flush-caches
```

Ordering matters: the snapshot (which protects pre-existing connections' current IPs) runs against the warm cache first, then the flush makes all future resolution visible. `resolvectl flush-caches`, or the D-Bus `org.freedesktop.resolve1.Manager.FlushCaches` call to avoid shelling out. Failure should be non-fatal (log and continue) — hosts without systemd-resolved already work via the redirect alone.

Cost: a one-time cold-cache burst at attach (tens of upstream queries, milliseconds each). No policy widening, no new map entries, full hostname attribution retained.

## Residual gap (accepted)

The flush covers everything resolving via the OS (python, curl, git, .NET on new connections). It does not cover a process that resolved pre-attach and holds the IP in app memory without an open socket (JVM-style in-process caching). If that shows up in practice, the escalation path is SNI-based verdicting on otherwise-blocked 443 flows — a much larger BPF change (hold verdict until ClientHello; SNI is forgeable by confined code so it needs pairing with IP plausibility). Not worth it until the flush proves insufficient.

## Rejected alternative: Azure Storage service-tag CIDR allowance

Auto-allowing `Storage.<region>` ranges on 443 when a `blob.core.windows.net` rule exists would also fix this (~250 prefixes for eastus+eastus2 against the 4096-entry LPM tries — capacity and per-packet cost are non-issues; LPM lookup is O(key bits)). It doesn't widen reachability vs the existing suffix rule, but it loses hostname attribution (flows become bare-IP allows), covers all Storage endpoints rather than just blob, and adds a weekly-freshness dependency on the service-tag file. The flush achieves the same outcome with none of those costs.

matthewdevenny:
Closes #89.

## Problem

On VM-style runners (reported case: a GitHub Actions runner on Azure, enforce mode), systemd-resolved's stub cache serves pre-attach answers **invisibly**. The DNS redirect intercepts resolved's *upstream* queries, but a stub cache hit at `127.0.0.53` never goes upstream — so the proxy never sees the name, no hostname↔IP mapping is learned, and the connection arrives as a bare IP → deny-by-default. The result was unattributed bare-IP blocks (e.g. the runner's results uploader to `productionresultssa*.blob.core.windows.net`) shipped to the SaaS as apparent policy violations. Whether a run was hit was a TTL lottery on whether resolved's cache happened to be warm.

Neither existing mitigation reaches this: the pre-existing-connection scan only covers sockets already open at attach, and the Phase-1 cache snapshot can only query *tracked* hostnames — a wildcard/suffix-matched name (`blob.core.windows.net`) is unknowable in advance, since the concrete storage-account name only exists once someone resolves it.

## Fix

Flush systemd-resolved's cache once the redirect is in place, so every subsequent lookup from any process misses the stub, goes upstream, and traverses the proxy — where suffix rules match, IPs are firewall-allowed, and attribution is kept.

Ordering (`cmd/start.go`):

```
install DNS redirect → snapshot tracked hostnames (warm cache) → resolvectl flush-caches
```

The snapshot and existing-connection reverse-DNS run against the warm cache **first** (they need the currently-in-use IPs), then the flush makes all future resolution visible. Gated on the redirect actually being installed (`dnsRedirectActive`) — with no proxy in the upstream path, flushing would only churn the cache. A full-cache flush is deliberate: a targeted flush is structurally impossible (the account name only exists post-resolution — see #89's rejected alternatives).

## Hardening (from `/code-review` + PR review)

- **Bounded** — the `resolvectl flush-caches` call runs under `exec.CommandContext` with a 5s deadline (threaded from the startup `ctx`), so a wedged systemd-resolved / D-Bus can't stall startup before the eBPF program attaches and enforcement begins.
- **Error classification** — only the genuine not-installed case (`errors.Is(err, exec.ErrNotFound)`) is swallowed; other `LookPath` failures (e.g. a non-executable `resolvectl` on PATH) surface to the caller instead of silently skipping the flush.
- **Quiet where expected** — when `resolvectl` is present but systemd-resolved isn't running (probed via `/run/systemd/resolve`), skip at `Debug`, not `Warn`, so hosts using a different resolver don't warn on every startup.
- **Tests** — new `pkg/network/dns_redirect_test.go` covers not-installed, not-running, flush-failure, success, and the timeout bound.

## Behavior notes

- **Non-fatal everywhere** — hosts without systemd-resolved already work via the redirect alone; a flush failure is logged and startup continues.
- **k8s sidecar: inert by design** — the cargowall container has no systemd-resolved, so both guards skip. The bug doesn't exist under cluster DNS anyway: pods resolve via a non-loopback ClusterIP / NodeLocal address, which the redirect DNATs straight to the proxy, so every query is already seen.
- **Residual gap (accepted, per #89)** — a process that resolved pre-attach and holds the IP in app memory with no open socket (JVM-style in-process caching) isn't covered; the escalation path is SNI-based verdicting, not worth it until the flush proves insufficient.

## Testing

**Unit** — on real linux (arm64): `go build ./...`, `go vet`, `staticcheck`, `make fmt-check`, and full `go test ./...` all green, including the 5 new `pkg/network` tests (the timeout test confirms the call is killed at its deadline).

**Integration (`ci.yml`, on the Azure runner where systemd-resolved is the real resolver — the #89 environment)**:
- *Enforce Mode* — `raw.githubusercontent.com` is warmed into the systemd-resolved stub cache **before** cargowall attaches, so the existing "allowed via `**.githubusercontent.com` pattern" curl doubles as a #89 regression check: if the attach-time flush regresses, the warm stub answer is served invisibly and the connection is blocked as a bare IP, failing the step.
- *GitLab CI mode* — asserts the `FlushResolvedCache` path actually executed at startup by grepping the captured debug log for the flush line (or the documented best-effort skip on hosts without systemd-resolved).

Both are additive and low-flake — the enforce check only fails if the flush is removed, and the gitlab-ci check reads a log the job already captures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,6 +140,22 @@ jobs:
       - name: Make binary executable
         run: chmod +x bin/cargowall
 
+      - name: Warm systemd-resolved cache before attach (reproduce #89)
+        run: |
+          # GitHub's Azure runners resolve via systemd-resolved (127.0.0.53).
+          # Populate its stub cache with a wildcard-matched name BEFORE cargowall
+          # attaches. The concrete name can't be pre-tracked, so only the
+          # attach-time cache flush makes the later getaddrinfo miss the stub and
+          # re-resolve through the proxy — where **.githubusercontent.com matches
+          # and the IP is allowed with attribution. Without the flush the warm
+          # entry is served invisibly and the "hostname pattern" curl below is
+          # blocked as a bare IP. Best-effort so runners without systemd-resolved
+          # simply fall back to the cold-cache path.
+          resolvectl query raw.githubusercontent.com 2>/dev/null \
+            || getent ahosts raw.githubusercontent.com \
+            || true
+          echo "Warmed resolver cache for raw.githubusercontent.com"
+
       - name: Start CargoWall via Action
         uses: code-cargo/cargowall-action@aa88f652e49b36a660add09311d077383dbc2ae7 # v1.3.3
         with:
@@ -159,8 +175,13 @@ jobs:
 
       - name: Test allowed connection via hostname pattern
         run: |
+          # Doubles as the #89 regression check: raw.githubusercontent.com was
+          # warmed into the systemd-resolved stub cache before attach (step
+          # above). If the attach-time cache flush regresses, this getaddrinfo
+          # is served invisibly from the warm stub, the proxy never sees the
+          # name, and the connection is blocked as a bare IP — failing here.
           curl -sf --connect-timeout 10 --max-time 30 https://raw.githubusercontent.com/github/gitignore/main/README.md > /dev/null
-          echo "✅ raw.githubusercontent.com allowed via **.githubusercontent.com pattern"
+          echo "✅ raw.githubusercontent.com allowed via **.githubusercontent.com pattern (warm-cache re-resolved through proxy)"
 
       - name: Test blocked connection
         run: |
@@ -732,6 +753,24 @@ jobs:
       - name: Show startup log (debug aid)
         if: always()
         run: sudo cat /tmp/cargowall.log || true
+
+      - name: Verify systemd-resolved cache flush ran at attach (#89)
+        run: |
+          # On the Azure runner the client resolver is systemd-resolved, whose
+          # warm stub cache used to serve pre-attach answers invisibly (#89).
+          # With the DNS redirect active (--gitlab-ci preset), cargowall flushes
+          # that cache at attach so future lookups traverse the proxy. Assert the
+          # FlushResolvedCache path executed — the flush itself on a resolved
+          # host, or the documented best-effort skip elsewhere.
+          if sudo grep -q "Flushed systemd-resolved DNS cache" /tmp/cargowall.log; then
+            echo "✅ resolved cache flushed at attach"
+          elif sudo grep -qE "systemd-resolved not running|resolvectl not found" /tmp/cargowall.log; then
+            echo "ℹ️ systemd-resolved not in use on this runner; flush correctly skipped"
+          else
+            echo "❌ neither flush nor skip logged — FlushResolvedCache path not reached"
+            sudo grep -iE "resolv|flush|dns redirect" /tmp/cargowall.log || true
+            exit 1
+          fi
 
       - name: Verify pidfile written
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -168,20 +168,24 @@ jobs:
           binary-path: ./bin/cargowall
           debug: true
 
+      - name: Test allowed connection via hostname pattern
+        run: |
+          # Runs FIRST among the connection tests: it doubles as the #89
+          # regression check for raw.githubusercontent.com, which was warmed
+          # into the systemd-resolved stub cache before attach (step above).
+          # Keeping it first minimizes the warm→request window so a short record
+          # TTL can't expire the entry and silently degrade this to the ordinary
+          # cold-cache path. If the attach-time cache flush regresses, this
+          # getaddrinfo is served invisibly from the warm stub, the proxy never
+          # sees the name, and the connection is blocked as a bare IP — failing
+          # here.
+          curl -sf --connect-timeout 10 --max-time 30 https://raw.githubusercontent.com/github/gitignore/main/README.md > /dev/null
+          echo "✅ raw.githubusercontent.com allowed via **.githubusercontent.com pattern (warm-cache re-resolved through proxy)"
+
       - name: Test allowed connection
         run: |
           curl -sf --connect-timeout 10 --max-time 30 https://github.com > /dev/null
           echo "✅ github.com allowed"
-
-      - name: Test allowed connection via hostname pattern
-        run: |
-          # Doubles as the #89 regression check: raw.githubusercontent.com was
-          # warmed into the systemd-resolved stub cache before attach (step
-          # above). If the attach-time cache flush regresses, this getaddrinfo
-          # is served invisibly from the warm stub, the proxy never sees the
-          # name, and the connection is blocked as a bare IP — failing here.
-          curl -sf --connect-timeout 10 --max-time 30 https://raw.githubusercontent.com/github/gitignore/main/README.md > /dev/null
-          echo "✅ raw.githubusercontent.com allowed via **.githubusercontent.com pattern (warm-cache re-resolved through proxy)"
 
       - name: Test blocked connection
         run: |
--- a/pkg/network/dns_redirect.go
+++ b/pkg/network/dns_redirect.go
@@ -56,6 +56,32 @@ func SetupDNSRedirect(logger *slog.Logger) error {
 	return nil
 }
 
+// FlushResolvedCache clears systemd-resolved's DNS cache via
+// `resolvectl flush-caches`. Once the DNS redirect is installed, flushing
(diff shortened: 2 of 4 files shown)


pkg/network/dns_redirect.go
```diff
@@ -56,6 +60,61 @@ func SetupDNSRedirect(logger *slog.Logger) error {
...
+			logger.Debug("resolvectl not found; skipping systemd-resolved cache flush")
+			return nil
+		}
+		return fmt.Errorf("locating resolvectl failed: %w", err)
+	}
+
+	// resolvectl can be installed on hosts that don't actually run
+	// systemd-resolved (a different resolver is in use). Flushing there only
+	// errors, so skip quietly rather than warn on every startup.
+	if _, err := os.Stat(resolvedRuntimeDir); err != nil {
```
jordanjennings, line 107: Any stat failure here — EACCES, EIO, not just absence — reads as "resolved not running" and skips at Debug, which the gitlab-ci grep then reports as a correct skip even on a host where resolved *is* running. Quiet-skip only `errors.Is(err, fs.ErrNotExist)` and surface the rest, matching the LookPath classification above.

pkg/network/dns_redirect.go
```diff
@@ -56,6 +60,61 @@ func SetupDNSRedirect(logger *slog.Logger) error {
...
+		logger.Debug("systemd-resolved not running; skipping cache flush", "probe", resolvedRuntimeDir)
+		return nil
+	}
+
+	flushCtx, cancel := context.WithTimeout(ctx, flushResolvedTimeout)
+	defer cancel()
+	if out, err := exec.CommandContext(flushCtx, path, "flush-caches").CombinedOutput(); err != nil {
```
jordanjennings: Without `WaitDelay`, the ctx kill signals only the direct child — `CombinedOutput`'s `Wait` still blocks until every holder of the internal stdout/stderr pipes exits, so the 5s bound rests on resolvectl never forking. Setting `cmd.WaitDelay = time.Second` makes it unconditional (it's also what `TestFlushResolvedCache_Timeout` implicitly depends on today, via `/bin/sh` exec'ing the fake's final command). While here: on timeout the returned error reads `signal: killed`; wrapping `flushCtx.Err()` would name the deadline in the log.

.github/workflows/ci.yml
```diff
@@ -159,8 +175,13 @@ jobs:
 
       - name: Test allowed connection via hostname pattern
         run: |
+          # Doubles as the #89 regression check: raw.githubusercontent.com was
+          # warmed into the systemd-resolved stub cache before attach (step
+          # above). If the attach-time cache flush regresses, this getaddrinfo
+          # is served invisibly from the warm stub, the proxy never sees the
+          # name, and the connection is blocked as a bare IP — failing here.
           curl -sf --connect-timeout 10 --max-time 30 https://raw.githubusercontent.com/github/gitignore/main/README.md > /dev/null
```
jordanjennings, line 182: This detects a regressed flush only while the warmed record is still in the stub cache — the action startup plus the github.com curl sit between the warm step and this request, so a short record TTL lets the entry expire and the check silently degrades to the pre-existing test (false-pass only, never a flake). If you want it tighter, run this curl first among the connection tests.

jordanjennings (approved):
LGTM with a few minor suggestions

pkg/network/dns_redirect.go
matthewdevenny: Good call — fixed in 120b1ff. The runtime-dir probe now quiet-skips only `errors.Is(err, os.ErrNotExist)`; EACCES/EIO and the like are surfaced as `probing %s failed: %w`, so a host where resolved *is* running can no longer be misreported as a correct skip. Mirrors the LookPath classification directly above it.

pkg/network/dns_redirect.go
matthewdevenny: Fixed in 120b1ff: set `flush.WaitDelay = time.Second` and, on deadline, return `flushCtx.Err()` wrapped so the log names the timeout instead of a bare `signal: killed`.

This also caught a latent test bug you implicitly predicted: the fake resolvectl was on a stripped `PATH`, so `/bin/sh` couldn't find `sleep` (exit 127) — the timeout path never actually ran, and the weak `err != nil` assertion passed anyway. The fake is now first on a full `PATH`, and `TestFlushResolvedCache_Timeout` asserts `errors.Is(err, context.DeadlineExceeded)` and runs ~1.05s — i.e. sh forks `sleep`, the child holds the pipes, and `WaitDelay` bounds the post-kill `Wait` (without it, `Wait` would block the full 5s). Nice catch.

.github/workflows/ci.yml
matthewdevenny: Done in 120b1ff — moved this curl to run first among the enforce connection tests, so the warm→request window is just the action startup with no intervening github.com fetch. Agreed the failure mode is false-pass-only (never a flake); running it first keeps the regression signal live under short record TTLs.

PR #95, opened 2026-07-29
#94 causal per-step attribution of network events (phase 0)

### LINKED ISSUE #94: Per-step causal attribution of network events (Phase 0)
Attribute every network event to the workflow step that (transitively) created its socket, replacing timestamp-based correlation.

Phase 0 scope (audit only, no enforcement changes):
- eBPF: tag Runner.Worker's direct children with a step ordinal at fork (tp_btf, separate collection so non-BTF kernels degrade attribution only); propagate tags across fork; bind tag to socket cookie at sock_create/connect; embed in TC events.
- Daemon: worker discovery (ancestry-first), pre-existing process seeding, reconciler emitting step_boundary audit events with the step process cmdline.
- Audit/OTLP: step_ordinal on connection events; step_scope for runner/pre-daemon sentinels.

## Rollout

Phase 0 (this issue) is attribution-only and severable. The follow-on work, in order:

1. **#103 — plan-based step naming**; retire the diag watcher and the `actions: read` permission. Cheap, independent, and ships a permission reduction on its own.
2. **#106 — container attribution** via root-cgroup egress. Moved ahead of per-step policy: it completes attribution before policy is built on it, and lands the final verdict hook so the policy data plane is only built once.
3. **#104 — per-step mode and default action.**
4. **#105 — per-step allowlists**, including the split between step traffic and CI-infrastructure traffic.

Design rationale for each lives in its issue. The ordering keeps one property from Phase 0: nothing that changes the verdict path ships before the attribution beneath it has run in production as audit-only.

matthewdevenny:
Closes #94.

## What

Attributes every network event to the workflow step that **causally** created its socket — not the step that happened to be running when packets flowed. Audit-only: no verdict consults the new maps, and any failure to start attribution degrades to a warning without touching enforcement.

## How

- **`bpf/stepbpf.c` (new collection):** `tp_btf/sched_process_fork` tags each direct child of Runner.Worker with a fresh step ordinal (synchronously inside fork, before the child's first instruction — no detection race) and every other fork inherits the forking thread's tag; `sched_process_exit` cleans up; `cgroup/sock_create` binds the tag to the socket cookie. Kept separate from `tcbpf.c` and loaded via `MapReplacements` so a kernel without BTF loses attribution only — the firewall loads untouched.
- **`bpf/tcbpf.c`:** shared step maps; the event struct's former pad word becomes `step_ordinal` (wire layout unchanged); TC embeds the socket's tag in every event; connect/sendmsg hooks re-tag as a pre-attach fallback.
- **`pkg/steps` (new):** worker discovery (ancestry walk first — a comm scan on a multi-runner host could lock onto another job's worker — falling back to a scan only when unambiguous), create-only seeding of pre-existing processes (never clobbers a kernel-assigned tag), and a reconciler that emits `step_boundary` audit events, retrying cmdline reads across the fork→exec window.
- **Audit/OTLP/summary:** `step_ordinal` on connection events end-to-end (including the RecentBlocks late-allow round-trip); sentinels export as `cargowall.step_scope` rather than bogus step ids; `step_boundary` rows are excluded from the summary's connection tables and the OTLP sink; the "no network events" path now works with boundary rows present.
- **Cost control:** step maps shrink to 64 entries when `--step-attribution` is off (it's implied by `--github-action` only), avoiding ~7MB of preallocated kernel memory for everyone else.

## Since the first push (review-driven)

- **Causal summary rendering:** events group under the step whose process created their socket (`#4 — Test blocked direct IP connection`), with GitHub names resolved by monotonic boundary-timestamp→step-index mapping (index, not name — names repeat); untagged traffic lands in labeled buckets (Docker-mediated / auto-allowed platform / unattributed); metrics show unique **and** raw event counts.
- **DNS blocks attributed:** the proxy resolves the querying step from its client socket via sock_diag (source-address dump; cookie parsing pinned against `SO_COOKIE`).
- **SaaS push is causal, no proto change:** same `steps[{name, events}]` schema, events assigned by ordinal→step-index instead of time window; ordinals stay agent-local.
- **Hardening from three review rounds + CodeQL:** boundary cmdlines emit argv[0..1] only (secrets can't ride flags into logs/JSONL/summary); markdown table cells escape pipes/newlines/backticks; bounded pid→u32 conversions; nil-objects guard in `steps.Start`.

## Testing

- Unit + root BPF suites in a Linux VM: fork/inherit/exit tagging against a live kernel (`TestStepFork`), BTF↔Go layout guards for both event structs (with BTF anchors added, since function-local types are dropped from BTF), shrunk-map load, ordinal-base validation, OTLP filtering/attrs, summary empty-path and boundary-exclusion.
- The enforce-mode integration job now **asserts** `step_boundary` events exist and that connection events carry real step ordinals — attribution runs against a genuine Runner.Worker in that job, and since the daemon deliberately warn-degrades, these assertions are what turn a silent degrade on real runner infrastructure into a red build.
- Empirical note baked into the design: kernel ordinals are opaque causal group IDs, not job-plan positions (Go's one-time pidfd probe fork consumes an ordinal, caught by the VM test) — plan correlation happens downstream via `step_boundary` cmdlines.

Follow-on work, in order: #103 (plan-based naming + `actions: read` removal), #106 (container attribution), #104 (per-step mode), #105 (per-step allowlists).


mtmk:
Review of the causal attribution path, roughly in severity order.

1. `resolveOrdinalSteps` makes the ordinal to step map injective (`prev = best`, `for i := prev + 1`), so N boundaries always consume N distinct plan steps. The kernel side mints one ordinal per direct-child fork of Runner.Worker, and a composite action's `run:` blocks are several forks under a single reported step. The second boundary inside one step finds no candidate (every remaining step started later), falls through to `prev + 1`, and from there every ordinal carries the wrong step name, in both the summary table and the SaaS push. No test covers two boundaries in one step; `TestSummary_ResolveOrdinalSteps_BoundaryBeforeRecordedStart` exercises the fallback only for the intended rounding case, which is indistinguishable from this one.

2. `sanitizeCmdline` keeps argv[0..1]. A standard `run:` step is `/usr/bin/bash -e /home/runner/work/_temp/<guid>.sh`, three fields, so every step reduces to `/usr/bin/bash -e ...` and the script path goes away. That path is the token the surrounding comments name as the correlation key. Node actions lose the action path the same way. `TestSanitizeCmdline` only asserts the two-field case, and the summary tests feed short strings that are already post-sanitization.

3. `buildCausalPushGroups` emits events in group order (scaffold first, then the runner/pre-daemon/docker buckets appended), but `pushToApi` still takes the job window from `allEvents[0]` and `allEvents[len-1]`. One pre-daemon block plus one later step block sends started_at after completed_at. This worked before only because temporal grouping happened to be chronological.

4. The causal branch skips `correlateEventsToSteps`, which also backfilled steps whose `completed_at` is null (from the next step's start, or maxEventTime+1s for the last one). GitHub returns null for the in-progress step, so those steps now push with no completion time. `--github-action` implies step attribution, so this is every real run.

5. `lookupSocketCookie` requires `src == want`. An autobound UDP socket (sendto with no connect, e.g. musl) has idiag_src 0.0.0.0 while the packet carries the routing-selected source, so the only matching row is rejected and the query is tagged ordinal 0, landing in Unattributed. Both sockdiag tests use connected sockets.

6. `causalGroups` builds `GitHubStep{Name: label}` with zero timestamps, so the heading's time range never renders even when the ordinal resolved to a real step. `ordinalSteps[ord]` already has the index to copy them from.

7. The blocked-DNS path now does a synchronous full socket-table netlink dump (500ms timeout) per query, on the handler goroutine, with no cache or concurrency bound. Under a query flood, which is the case filtering exists for, that competes with legitimate resolution on the same proxy. Reasoned from the code, not measured.

8. `sanitizeCmdline` and `readCmdline` write U+2026 into the audit JSONL and the rendered summary, group labels use an em dash that the test then has to `strings.Cut` on, and the added comments use arrows and dashes throughout. Repo rule is ASCII only. Separately `readCmdline` truncates at `s[:256]` on bytes, which can cut a rune mid-sequence.

9. `SetStepLookup` was inserted between `SetRecentBlocks`'s doc comment and its declaration, so the comment now documents the wrong function and `SetRecentBlocks` has none.

10. `lookupSocketCookie` collapses socket/sendto/parse/NLMSG_ERROR/timeout failures into `return 0, false` with no logging, so a kernel without CONFIG_INET_DIAG is indistinguishable from a genuine no-match. Tracker already carries a logger.

mtmk (approved):
LGTM

Issue #106, opened 2026-08-07
Container attribution via root-cgroup egress


matthewdevenny:
Rollout order after #94 (attribution, PR #95): #103 → #106 → #104 → #105. Moved ahead of #104/#105 — see below.

Closes the last attribution gap: traffic created by containers and the Docker daemon, which today lands in the summary's "Docker-mediated" bucket because it can't be tied to the step that caused it.

## Why this comes before per-step policy

Originally sequenced last; moved ahead of #104/#105 for two reasons:

1. **Otherwise it forces rework in the riskiest code.** Per-step policy means a per-policy lookup in the verdict path. Landing that on TC egress first, then migrating the hook here, means porting the whole per-step lookup to a different hook afterwards. Building the policy data plane once, on the final hook, is cheaper and safer.
2. **Per-step policy on incomplete attribution misleads.** Any container-using workflow has a large class of traffic that can't be tied to a step. With the fail-closed rule (untagged ⇒ strictest applicable policy), that either breaks container workloads or forces the daemon tier wide open — which is the flat host-wide allowlist #105 exists to eliminate. Users would believe their container steps are constrained when they are not.

In cargowall's own enforce-mode integration job — which deliberately exercises Docker — more destinations currently land in the untagged buckets than under real steps.

## Why it's the big one

Container packets cross a network namespace and NAT before the current TC egress hook sees them, so the socket is no longer available at that point — the cookie→tag lookup that powers attribution can't work there **even if the container's sockets were tagged perfectly**. The read side therefore needs a hook with socket context.

## Staging

Phase 0's safety property was: never change the verdict path until the attribution beneath it has run in production. Same split here.

### 3a — spike + container attribution (audit-only)

Open question to answer first: **can container traffic be attributed without moving the verdict?** Candidate is `cgroup_skb/egress` attached at the root cgroup as an observer (always allow, emit tagged records), with TC still enforcing, and userspace joining the two. If that works, the attribution story completes with zero enforcement risk.

Also in 3a:
- Tag container processes. Their ancestry runs through containerd-shim rather than the step process, so fork inheritance can't reach them. Subscribe to Docker events, resolve the workload PID (`ExecInspect` for job-container steps, container init for step-created containers), confirm identity from the process environment, then tag.
- Map the embedded resolver's forwarded queries so container DNS is attributable.
- Unattributable container traffic gets a dedicated tier — never the loosest step policy.

Measure during 3a, since the observer build gives the numbers for free:
[...]

PR #109, opened 2026-08-07
#106 container attribution + cgroup egress enforcement (phases 3a+3b)

matthewdevenny:
Closes the container attribution gap (#106) and, behind a flag, moves primary egress enforcement into socket context. Bridge-networked container traffic crosses a netns + NAT before TC sees it, so the cookie join there finds nothing. A root-cgroup `cgroup_skb/egress` hook sees the flow pre-NAT: it records who owned it and — once raised out of observe mode — decides its fate.

## Mode ladder

The hook attaches early but stays inert; the mode is raised only after the allowlist is programmed.

- **observe** — pass everything, record flow origins (phase 3a).
- **shadow** — default with `--container-attribution`: compute verdicts, report `cgroup_would_block`, block nothing. Measures the blast radius before anyone relies on it.
- **enforce** — `--cgroup-enforce`: drop denied traffic at the hook. TC egress stays attached as the fail-closed backstop. Startup refuses the flag if its prerequisites (`--container-attribution`, `--step-attribution`) are missing — a requested enforcement posture is never silently downgraded.

## Design in two sentences

The allow/deny decision lives once, in `bpf/verdict.h`, shared by both hooks through `MapReplacements` — a parity test proves TC and the cgroup hook cannot disagree. Both hooks report through one post-verdict pipeline (hostname resolution, late-allow self-healing, audit record, notification), so an outcome reads the same regardless of which hook produced it.

## Attribution

- `pkg/origin` keeps a bounded join store of pre-NAT origins. TC events with no socket identity are enriched from it: container id, plus only the socket's own step tag — never "current step" — so untagged traffic lands in the stricter container tier. Ambiguity always degrades stricter: source-port collisions across netns return every candidate, and PID disagreement drops the process claim.
- `pkg/containers` (decomposed: tracker / reconcile / start_exec / network / enricher) tracks docker and network events, verifies PID identity via cgroup before tagging — a refused identity is never tagged, not even by a replayed start event — re-tags on `docker exec`, survives daemon flaps (two-pass stale sweep, incarnation re-adoption, duration-based backoff), and attributes container DNS by client IP.
- The summary gains a container-unattributed tier; OTLP carries the container fields.

## Safety properties, each pinned by a test

- Non-enforcing modes pass every packet, however malformed.
- Audit posture ("log, never block") holds at this hook exactly as at TC.
- Every enforce-mode drop of a connection attempt or live flow emits a record; a verdict change re-emits through the dedup interval, so a drop can never be invisible to audit.
- Local-only traffic (loopback addresses, and anything egressing `lo` — including the host's own IP) and ICMP are carved out: the hook never polices what TC never covered. Docker bridge subnets are the same class, carved via a map only this hook reads (`map_local_nets`, never shared with TC — a workload-influenced subnet must not widen off-host enforcement, and a test proves the isolation by actually sharing the verdict maps). Coverage: networks are enumerated directly (`GET /networks`) before the mode is raised, on network-create events thereafter, and per-container as backstop — so published ports, container→container, and networks that predate cargowall all survive enforcement. CI exercises every shape in shadow and enforce.
- The carve-out write validates its own input (`origin.AllowLocalNetwork`): bridge-driver networks only, width-capped (/16 v4, /64 v6), refused when the subnet contains a real host interface address (the one exemption — the subnet's own gateway — must sit on a genuine linux bridge, so a tun device claiming its range is refused). Fail-closed on enumeration errors; refusals are a distinct error (`ErrLocalNetworkRefused`) so callers can tell a decision from a transient failure.
- Under audit posture, the cgroup hook's suppressed denials report exactly like TC's (normalized `connection_blocked` with late-allow), never as shadow telemetry that summary/OTLP discard.
- Verdict reporting survives saturation: overflowed Blocks divert to a bounded must-audit lane (bare-IP audit record, no PTR/notification); a report is lost only when both lanes are full, and the Close summary splits `dropped` / `degraded` / `abandoned` so runtime saturation is never confused with shutdown bookkeeping.
- DNS attribution is a constructor-time property of each listen address (host-sockdiag vs container-IP) — bridge queries never touch sockdiag, and running interception without attribution files them as unattributed container traffic (a pinned, accepted residual).
- The DNS proxy's marked traffic is exempt, so a policy race can't self-block resolution.
- Verdict reporting is async and store-first: a slow PTR lookup can neither lose ring records nor starve the join store TC enrichment reads.

Behavior change worth knowing outside enforcement: `ensureAllowed` (DNS/infra auto-allows) now accepts CIDR prefixes and IPv6 instead of silently discarding them — so a systemd-resolved IPv6 upstream or a v6 `--dns-upstream` gets its port-53 allow rule where it previously got nothing.

## Verified

Kernel tests (`make test-bpf`): TC↔cgroup verdict parity, mode/audit/carve-out/emission semantics, the local-nets isolation property (against genuinely shared maps), BTF layout pin, real-socket cookie join, and a live-socket verdict-change re-emit. Full linux unit suite plus `-race` on the touched packages, including tables for the subnet-validation policy, the deny-veto rules, and end-to-end fake-daemon coverage of the identity gate. CI gates the shadow-mode blast radius (zero would-blocks on loopback, bridge, or allowed hosts) before any enforce job runs, and the enforce job proves published-port, default-bridge, live-created, and pre-existing bridge networks all work under enforcement.

Heads-up for audit-log readers: the BPF suite's enforcement probe targets the host's own IP over `lo`, so a production cargowall on the same runner never sees it — no `bpf.test` lines in real audit streams or product reports. The one deliberate off-host denial (`203.0.113.9:9999`, TEST-NET-3) lives in the offline cgroup-enforce CI job, where the policy is local and the block is asserted in-job and reported nowhere else.


bpf/originbpf.c
```diff
@@ -0,0 +1,572 @@
...
+        dst_port = bpf_ntohs(udp_hdr.dest);
+    }
+
+    __u8 allowed = 1;
+    __u8 verdict = ORIGIN_VERDICT_NONE;
+    if (mode != ORIGIN_MODE_OBSERVE) {
+        // Carve-out evaluated only when a verdict would be, so observe mode
+        // never pays for it; operand order keeps the config-map lookup
+        // confined to packets actually on the loopback device.
+        __u8 carve_out = ((dst_ip >> 24) == 127) || (ip_proto == IPPROTO_ICMP) ||
```
jordanjennings: The carve-out covers 127/8, ICMP, and `lo` by ifindex, but not the docker bridge. Under `--cgroup-enforce` that makes two traffic shapes deny-by-default, neither of which TC ever adjudicated:

**Host → published container port.** `docker run -p 8080:80` + `curl localhost:8080`: docker's nat/OUTPUT rule excludes 127/8, so docker-proxy handles it and opens its own socket to 172.17.0.2:80. That socket's egress reaches here with dst 172.17.0.2 on docker0 — no carve-out matches, and no allowlist rule covers 172.17.0.0/16 (the only auto-adds are 127.0.0.0/8 + ::1/128 at `cmd/start.go:446`, the CI-only DNS IPs at `:1179`, and cloud metadata). `verdict_allowed_v4` falls to default-deny and the SYN dies in `ip_finish_output`.

**Container → container** on a user-defined bridge, adjudicated in the container's netns.

The shadow gate at `.github/workflows/ci.yml:419` is the right instrument — it asserts zero would-blocks on `172.17.*` — but it passes vacuously: no CI job publishes a port or runs container-to-container traffic, so neither shape is ever generated. The only bridge traffic exercised is container→gateway DNS, which is allowed explicitly.

Adding the bridge CIDR to the carve-out (or to the allowlist), plus a shadow-job step that publishes a port and curls it, would give the existing `172.17` assertion traffic to actually assert on.

bpf/originbpf.c
```diff
@@ -0,0 +1,572 @@
...
+    key.dst_port = dst_port;
+    key.ip_proto = ip_proto;
+    key.ip_version = ip_version;
+
+    __u64 now = bpf_ktime_get_ns();
+    struct origin_seen_val *last = bpf_map_lookup_elem(&map_origin_seen, &key);
+    if (last && last->verdict == verdict && now - last->last_emit < ORIGIN_EMIT_INTERVAL_NS)
+        return;
+    struct origin_seen_val val = { .last_emit = now, .verdict = verdict };
+    bpf_map_update_elem(&map_origin_seen, &key, &val, BPF_ANY);
```
jordanjennings: The dedup entry is stamped before the ring reservation, so a failed reserve loses the record *and* then suppresses re-emission for the full 10s `ORIGIN_EMIT_INTERVAL_NS`.

The comment above says a reservation failure means "attribution degrades, nothing else does" — true in observe/shadow, where TC still emits its own event. In enforce mode it isn't: for traffic that dies here (container traffic, dropped pre-NAT in its own netns) this hook is the sole event source, so the packet is dropped with no `connection_blocked` record, no notification, and no late-allow self-healing — and every retransmit for the next 10s matches the stamped entry and returns early. That contradicts the PR's "every drop we cause emits a record".

Moving the `bpf_map_update_elem` below the `if (!evt) return;` fixes it: a lost record then re-emits on the next packet instead of going quiet.

bpf/originbpf.c
```diff
@@ -0,0 +1,572 @@
...
+            // replies to inbound port scans and SYN-ACK replies to inbound
+            // handshakes (TCP with flags 0 — constant noise on a public CI
+            // runner), and non-first fragments (the flow's first packet
+            // already carried this verdict).
+            if (!allowed && !non_first_frag && (ip_proto != IPPROTO_TCP || flags != 0))
+                emit = 1;
+        }
+    }
+
+    if (emit) {
```
jordanjennings, line 509: `emit` is initialised to 1 at line 332 and the carve-out only skips the verdict, so carved-out v4 traffic still emits: every loopback TCP SYN, every loopback UDP datagram, and every ICMP packet produces a record that nothing can ever join, since none of it reaches TC.

The DNS redirect DNATs port 53 to 127.0.0.1 before this hook runs, and each query uses a fresh ephemeral socket — hence a fresh cookie and a fresh `map_origin_seen` key — so this is one record per host DNS query in a 128KB ring, plus the reader-side map work in `Observer.insert` and a dead key in the join store.

The v6 handler already does the right thing: line 428 early-returns on multicast and `::1` without emitting. Setting `emit = 0` in this carve-out branch would make the two handlers consistent.

cmd/flags.go
```diff
@@ -205,6 +217,23 @@ func (c *StartCmd) AfterApply() error {
...
+	// nil and the requested posture evaporates. (Runs after preset
+	// expansion, so --github-action — which implies both — satisfies the
+	// requirement.) Step attribution can still fail at RUNTIME (no kernel
+	// BTF, no Runner.Worker); that residual downgrade is warned loudly in
+	// newContainerAttribution.
+	if c.CgroupEnforce && !c.ContainerAttribution {
+		return fmt.Errorf("--cgroup-enforce requires --container-attribution")
+	}
+	if c.CgroupEnforce && !c.StepAttribution {
+		return fmt.Errorf("--cgroup-enforce requires --step-attribution (container attribution rides on step attribution)")
```
jordanjennings, line 235: `AfterApply` requires `--container-attribution` and `--step-attribution` for `--cgroup-enforce`, but not a CI preset — so `--container-attribution --step-attribution --cgroup-enforce --docker-dns-interception` is accepted.

On that path `cmd.DockerDNSInterception` still points daemon.json at the bridge IP (`cmd/start.go:277-303`), but the allow rule for it lives in `loadCIConfig` (`cmd/start.go:1176-1179`), which only runs when `CIMode() != CIModeNone` (`:399`). Every container DNS query then egresses to 172.17.0.1:53, matches no rule, and dies at the cgroup hook with EPERM — container name resolution fails completely, with `connection_blocked` events on port 53 as the only clue.

The `dnsIPs` block itself is pre-existing and unchanged here; the enforce hook is what makes it load-bearing. Gating the bridge allow on the same condition that installed the redirect, rather than on CI mode, is the cheaper fix than extending validation.

pkg/origin/origin.go
```diff
@@ -0,0 +1,597 @@
...
+	o.mu.Unlock()
+
+	// Report AFTER the store write above — see the ordering note in the doc
+	// comment. Non-blocking: a full queue costs the report, never the reader.
+	if rec.Verdict == VerdictWouldBlock || rec.Verdict == VerdictBlock {
+		o.blocked.Add(1)
+		select {
+		case o.verdictCh <- rec:
+		default:
+			if n := o.verdictsDropped.Add(1); n == 1 || n%1000 == 0 {
```
jordanjennings: The non-blocking send and the abandon-at-Close behavior are deliberate, documented at `:151-161`, and pinned by `TestInsertNeverBlocksOnFullVerdictQueue` — no argument with the tradeoff itself.

The issue is that the PR description and design.md state it more strongly than the code delivers: "every enforce-mode drop … a drop can never be invisible to audit" holds for the BPF ring, but a record whose *report* is dropped here produces no audit event, no notification, and no `tryLateAllow` reconciliation. In enforce mode, where this hook is the sole event source for what it drops, that is a silent enforcement action.

Either soften the claim to best-effort-under-overflow, or emit a degraded record inline on the `default:` arm (no PTR, no late-allow — bounded and cheap) so an enforce-mode drop is never event-less. Surfacing `verdictsDropped` in the run summary alongside the existing WARN would also help; right now the gap is only visible in logs.

pkg/containers/tracker.go
```diff
@@ -0,0 +1,793 @@
...
+		if ctx.Err() != nil {
+			return
+		}
+		t.logger.Warn("Docker event stream interrupted; reconnecting", "error", err, "backoff", backoff)
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(backoff):
+		}
+		backoff = min(backoff*2, maxBackoff)
```
jordanjennings, line 298: `backoff` is declared once at `:258` and only ever doubles — nothing resets it after a stream that ran successfully. A few early transient errors pin it at `maxBackoff` for the process lifetime, so a later dropped connection costs a 5s event-stream gap instead of 500ms.

The gap isn't self-healing. `reconcile` runs at `:283` before the decode loop, so a container born during the gap is adopted as `kind == "reconcile"` → `StepOrdinalPreDaemon` (`:486-487`) and `TagContainerProcess` is called with that value; the replayed `start` event then hits the `known` early-return at `:446-450`, so the correct event-time ordinal never reaches the init PID. Replayed `exec_start` events do get `OrdinalAt(event time)` and repair `effectiveOrdinal`, so the damage is confined to containers whose init subtree is their only attribution.

Resetting `backoff = reconnectBackoff` after a stream that delivered at least one event keeps the 500ms floor against a hot reconnect loop.

pkg/containers/tracker.go
```diff
@@ -0,0 +1,793 @@
...
+
+	if info.privileged {
+		t.logger.Warn("Privileged container started (host-root equivalent; observer coverage not guaranteed)",
+			"container", shortID(id))
+	}
+
+	t.mu.Lock()
+	t.containers[insp.ID] = info
+	for _, ip := range info.ips {
+		t.byIP[ip] = info
```
jordanjennings: Two *live* containers can't share an address under a single dockerd (libnetwork IPAM won't hand it out twice, and overlapping subnets are rejected at network create), so that isn't the real exposure here.

The reachable one is address recycling: the allocator reuses the lowest free ordinal, and `handleStart` only indexes after an inspect RTT plus a `/proc` cgroup read — and early-returns entirely on inspect error or `Pid == 0` (`:455-461`). In that window `byIP` still answers with the dead container, and `LookupClient` (`enrich.go:213-221`) returns its id **and** its `effectiveOrdinal` with `ok = true` — a wrong step stamped as fact, against the "untagged … the unattributed tier, never a step" promise at `:38-41`.

`Enrich` isn't affected (the ordinal comes from the origin record's own socket tag, `enrich.go:162-164`) and `DecorateVerdict` stamps id only, so this is audit-only. Detecting the overwrite here (`prev := t.byIP[ip]; prev != nil && prev != info`) and degrading that address to `ok=false` would fit the codebase's existing ambiguity vocabulary better than silently picking a winner.

pkg/containers/tracker.go
```diff
@@ -0,0 +1,793 @@
...
+// markExecSeen records execID under its container and reports whether it
+// had already been handled. Per-container keying exists so remove() can
+// reclaim the whole set when the container goes away.
+func (t *Tracker) markExecSeen(containerID, execID string) (seen bool) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	set := t.seenExecs[containerID]
+	if set == nil {
+		set = make(map[string]bool)
+		t.seenExecs[containerID] = set
```
jordanjennings: `handleExecStart` runs on its own goroutine (`:313-323`) and can still be parked in `resolveExecPid` — up to ~1s under `execSem`, or ~5s on the no-`execID` fallback — when the container's die/destroy reaches `remove` on the stream goroutine and deletes `t.seenExecs[id]`. `t.mu` makes each map operation atomic but imposes no ordering between them, so this then reinstalls the entry after teardown.

`remove` is the only deleter, and reconcile's stale sweep iterates `t.containers` (`:425-429`) — which a resurrected entry isn't in — so it's never collected.

Small (~300-500 bytes per *raced* container, not per container) and harmless beyond memory, since exec ids are unique and a stale set can't produce a false `errExecAlreadySeen`. Pruning orphans in reconcile's sweep is the cheap fix; gating `markExecSeen` on `t.containers` membership instead would break the case `:702-703` deliberately supports.

pkg/containers/tracker.go
```diff
@@ -0,0 +1,793 @@
...
+	}
+
+	t := &Tracker{
+		client:      newDockerClient(opts.Socket),
+		tagger:      tagger,
+		auditLogger: auditLogger,
+		logger:      logger,
+		opts:        opts,
+		done:        make(chan struct{}),
+		execSem:     make(chan struct{}, 4),
```
jordanjennings: `reconcileInspectConcurrency` at `:188` documents itself as "Same shape and rationale as execSem", but execSem's own bound is a bare literal here — so the cross-reference points at nothing greppable, and changing one leaves the other silently behind.

jordanjennings (changes requested):
Requesting changes on the `--cgroup-enforce` path. The `bpf/originbpf.c` carve-out omits the docker bridge, which makes published container ports and container-to-container traffic deny-by-default under the flag; the shadow gate that would have caught it has no traffic to assert on. Two smaller enforce-path issues alongside it: container DNS breaks when the flag is used without a CI preset, and a ring-reservation failure can suppress a drop's only audit record for 10s. The remaining inline comments are non-blocking.

Note on the description: the v6-guard removal in `ensureAllowed` means DNS/infra auto-allows now program IPv6 rules where they previously discarded them — a systemd-resolved v6 upstream, or a v6 `--dns-upstream`, gets a /128 port-53 rule it didn't get before. That reads as a fix rather than a regression, but it's a behavior change the description doesn't mention.

bpf/originbpf.c
jordanjennings: Two things in the repo's own docs that make this more concrete than "an unhandled case".

`design.md:556-559` names both surfaces in one breath, then supplies a remedy for one of them:

> The hook sees traffic TC never did — loopback (including the DNS proxy's own DNAT'd `127.0.0.1:53`) and the docker bridge. Loopback is carved out both in BPF and as a userspace `127.0.0.0/8` + `::1/128` allow rule; ICMP/ICMPv6 and IPv6 multicast are passed for PMTU and NDP.

And `pkg/dns/server.go:374-380` shows the same failure mode was already anticipated for one specific case — a local socket egressing to an address no allowlist would name:

> Replies to a DNS client are egress from a local socket addressed to that client's ephemeral port — which no allowlist would ever name — so once the cgroup egress hook is enforcing (issue #106 phase 3b) an unmarked reply is dropped and every client silently fails to resolve.

docker-proxy's socket to a container IP is that same shape, and gets no equivalent treatment. The DNS listeners were fixed because CI exercises them; the published-port path isn't exercised, so it wasn't.

One correction to my comment above, for whoever picks this up: `GetDockerBridgeIP` (`pkg/network/docker.go`) already holds the `net.Addr` and discards the mask, so the bridge *subnet* isn't computed anywhere today — returning the `*net.IPNet` is the prerequisite for the allowlist route. That also argues for fixing this in the rule maps rather than the BPF carve-out, since user-defined networks use arbitrary subnets that aren't compile-time constants.

matthewdevenny: Fixed in 4230bbc — but via a third route, not the two suggested, because your follow-up correction pointed at the real constraint: user-defined subnets are workload-influenced input. Putting them in the shared rule maps would let `docker network create --subnet 10.0.0.0/8` (or a macvlan network whose subnet is the physical LAN) widen **TC**'s off-host enforcement for every process on the host, permanently (rules are append-only and deny rules don't suppress infra auto-allows, as you noted).

Instead the carve-out lives in `map_local_nets`/`_v6` — LPM maps **owned by the origin collection**, never wired into `MapReplacements` — checked alongside the loopback carve-out. A hostile/overclaimed subnet can at most return local traffic to its pre-3b posture (unpoliced by a hook that didn't exist); anything leaving the host still meets TC's untouched verdict. `TestOriginEgressLocalNetworkCarveOut` pins both halves: pass at the cgroup hook, still dropped at TC. Discovery is also driver-gated (`/networks/{name}` inspect, bridge only) so macvlan/ipvlan space stays adjudicated.

Coverage: default bridge + pre-existing containers' bridges are carved before `enableMode` (`preallowLocalNetworks`, using the new `GetDockerBridgeSubnet` you noted was missing); later networks via tracker discovery. CI now generates both shapes — the shadow job publishes a port and runs container→container so the 172.17 gate finally has traffic, and the enforce job asserts published-port, default-bridge c2c, a network created *while* cargowall runs, and one created *before* it starts.

bpf/originbpf.c
matthewdevenny: Fixed in 4230bbc: the stamp moved below the reserve, so a failed reservation costs one packet instead of a 10s-quiet enforce drop. The header comment now owns the trade explicitly — the unstamped window spans the reserve, so N CPUs racing one flow can emit up to N duplicates (userspace dedups by cookie in the join store; a duplicate verdict record costs a duplicate audit line), and a persistently full ring retries per packet rather than going quiet.

bpf/originbpf.c
matthewdevenny: Fixed for 127/8 in 4230bbc: loopback destinations never emit now (`emit = (dst_ip >> 24) != 127`), matching the v6 handler's `::1` early return — which also killed the per-DNS-query dead record you called out, and forced the real-socket join test onto the host's own IP.

ICMP deliberately still emits, though: a denied *external* ICMP packet does reach TC as a protocol-block candidate (`submit_protocol_block_v4`), and that TC event joins against the origin record by dst+ports-0. So ICMP records aren't dead — only loopback's were.

cmd/flags.go
matthewdevenny: Fixed in 4230bbc: the DNS-infra allow (127.0.0.1, upstream host, bridge listener) moved out of `loadCIConfig` into the main startup path after config load, unconditional — gated on nothing, since even with tracking disabled processes still resolve against those addresses and the port-53 flows must survive default-deny. `loadCIConfig`'s now-unused `dockerBridgeIP` param is gone too.

pkg/origin/origin.go
matthewdevenny: Went with the stronger option in 86f0ec1/4230bbc: a bounded **must-audit lane**. When the main queue overflows, Blocks divert to a second small queue drained by its own worker, delivered with `Record.Degraded` — `ReportVerdict` then does bounded work only (no PTR, no late-allow, no notification; audit record with the bare IP, decoration intact, same kind switch as the full path). So an enforce-mode drop is event-less only when BOTH lanes are saturated. Close triages the backlog the same way: queued Blocks flush through the lane, would-blocks are abandoned.

The counters split accordingly and land in the Close summary: `verdict_reports_dropped` now only ever means runtime saturation, `_degraded` counts bare-IP audit records, `_abandoned_at_close` is shutdown policy. Unit tests pin the bounded contract and the two-lane overflow semantics.

pkg/containers/tracker.go
matthewdevenny: Fixed in 4230bbc — and your inclusive-`since` observation killed the obvious fix: an event-based reset is satisfied by the replayed event on every session that connects, exactly the flapping case backoff exists for. The health signal is duration-based instead (`streamHealthyAfter` = 30s of stream survival resets to the 500ms floor), which the replay cannot forge.

pkg/containers/tracker.go
matthewdevenny: Reworked in 4230bbc per this: the wholesale sweep is gone. A conflicted address now resolves newest-inspect-wins — right for IPAM recycling (the newcomer genuinely holds the address) and identical to the pre-change last-writer semantics for the manually-assigned cases where IPAM guarantees nothing. The displaced holder stays tracked (its cgroup identity and other addresses may be live); if it is truly dead, reconcile's stale sweep collects it. And the conflict log moved out of the critical section — displaced entries are collected under `t.mu` and logged after unlock, matching the reconcile pattern you pointed at.

pkg/containers/tracker.go
matthewdevenny: Fixed in 4230bbc (86f0ec1 had the first half): reconcile's sweep now also prunes `seenExecs` sets whose container is neither tracked nor in the live list — the orphan a parked exec handler can resurrect after `remove()`. Gating `markExecSeen` on membership was avoided for exactly the reason you flagged.

pkg/containers/tracker.go
matthewdevenny: Done in 4230bbc: `execConcurrency = 4` is a named const, `execSem` is built from it, and `reconcileInspectConcurrency`'s comment now references something greppable.

pkg/origin/origin.go
```diff
@@ -0,0 +1,353 @@
...
+	"net/netip"
+	"os"
+	"sync"
+	"sync/atomic"
+	"unsafe"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/link"
+	"github.com/cilium/ebpf/ringbuf"
+
```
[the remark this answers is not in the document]
matthewdevenny: No flag is needed here, but the reasoning deserved to be written down — done at the attach site. `link.AttachCgroup` attaches via **bpf_link** (`newLinkCgroup`) on kernels ≥5.7, under our 5.8 floor, and kernel cgroup bpf_links always attach with allow-multi semantics — coexistence is implied by the mechanism, not a flag. cilium/ebpf's legacy `PROG_ATTACH` fallback (unreachable above the floor) also tries `flagAllowMulti` first, `flagAllowOverride` last. Verified against cilium/ebpf v0.21.0's `link/cgroup.go`; the pre/post attach program-ID query logging is the observational check that no peer program was displaced.

matthewdevenny:
All review findings addressed in `4230bbc` (with some groundwork in `86f0ec1`) — per-thread details in the inline replies. The short version:

**Blockers**
- **Bridge traffic under `--cgroup-enforce`**: fixed via a cgroup-hook-only carve-out map (`map_local_nets`), deliberately *not* the shared rule maps — subnet values are workload-influenced, and a rule-map entry would let `docker network create --subnet 10.0.0.0/8` (or a macvlan LAN subnet) widen TC's off-host enforcement. Driver-gated to bridge networks; default bridge + pre-existing containers carved before `enableMode`; a test pins that the carve-out never leaks to TC. CI now exercises published-port and container→container in both the shadow gate and the enforce job, including a network created before cargowall starts.
- **Container DNS without a CI preset**: DNS-infra allows moved out of the CI-only path, now unconditional.
- **Ring-reservation muting**: dedup stamps only after a successful reserve.

**Also taken**: loopback records no longer emit (ICMP kept — denied external ICMP joins TC protocol-blocks); enforce drops survive reporter saturation via a bounded must-audit lane with split counters; duration-based backoff reset (the inclusive-`since` replay defeats event-based); byIP conflicts resolve newest-wins with no wholesale untracking and no logging under the lock; seenExecs orphans pruned in reconcile; `execConcurrency` named.

**On the review-body note**: the v6/CIDR `ensureAllowed` behavior change is now called out in the PR description, and the `GetDockerBridgeSubnet` prerequisite from the follow-up comment exists (deduped with `GetDockerBridgeIP` over one interface walk).

bpf/originbpf_test.go
```diff
@@ -0,0 +1,938 @@
...
+
+// The local-networks carve-out (map_local_nets): a subnet written there is
+// exempt from THIS hook's adjudication — and, critically, from this hook
+// ONLY. The map is owned by the origin collection and never shared, so a
+// workload-influenced subnet (docker bridge discovery) can exempt local
+// traffic this hook alone sees but can never widen tc_egress's off-host
+// enforcement; the parity test's premise (shared decision) is about the
+// verdict maps, which this deliberately is not part of.
+func TestOriginEgressLocalNetworkCarveOut(t *testing.T) {
+	objs := loadOriginObjects(t)
```
jordanjennings: This test's isolation half can't fail. `loadOriginObjects` (`:68-75`) passes `nil` collection options, so `objs` and `tcObjs` are two independently loaded collections sharing no kernel maps — the test itself has to program `tcObjs.MapDefaultAction` separately at `:455` after `seedOriginRules` already programmed origin's. The write at `:438-442` goes to origin's own `map_local_nets`; the assertion at `:456-464` then observes only that a deny-all TC drops `10.99.7.3`, which is true regardless of anything the carve-out did. It would stay green under exactly the regression it advertises catching — `map_local_nets` moved into `verdict.h` and consulted by `tc_egress`.

This matters more than a normal test nit because the PR comment and `design.md` both cite it as the thing that makes workload-influenced subnet discovery safe to wire ("a test pins that the carve-out never leaks to TC"). Nothing in the repo currently pins that `map_local_nets` stays out of `MapReplacements`.

`loadOriginObjectsSharing` (`:797-812`) already does the right wiring and is used by `TestVerdictParityWithTcEgress`. Reordering to `tcObjs := loadBPFObjects(t); objs := loadOriginObjectsSharing(t, tcObjs)` makes the existing assertion a real detector with no other change.

pkg/containers/start_exec.go
```diff
@@ -0,0 +1,382 @@
...
+	if !identityOK {
+		t.identityRejected.Add(1)
+		t.logger.Warn("Container identity check failed; not tagging", "container", shortID(id), "error", err)
+	}
+	info.cgroupID = cgroupID
+
+	var ordinal uint32
+	if kind == "reconcile" {
+		ordinal = events.StepOrdinalPreDaemon
+		info.preDaemon = true
```
jordanjennings: `preDaemon` is set without consulting `identityOK`, which lets a container whose cgroup identity check was refused get tagged later anyway.

The chain: `identityOK` is computed at `:83` and correctly gates tagging at `:110`, but registration at `:147` (`t.containers[insp.ID] = info`) is unconditional — only the `byCgroup` insert is gated. So a refused container sits in `t.containers` with `preDaemon == true`. When its real `start` event is replayed after any stream reconnect, the guard at `:43` (`kind == "start" && timeNano > 0 && existing.preDaemon`) fires, and `upgradeReconciledStart:199` calls `TagContainerProcess` with no re-verification, under the comment at `:198` — "Identity was verified at adoption; the PID has not changed since" — which is false in precisely this case.

Worst case is an unrelated host process carrying a real step ordinal for its lifetime, mis-credited in the audit log, summary and OTLP, with a `container_attribution` event asserting attribution that `:105-108` says is refused. Bounded: nothing keys a verdict on the ordinal (`tcbpf.c` only stamps `evt->step_ordinal`), so there's no enforcement impact; and where the check failed because the PID was already gone, the tag is a no-op since `tagTasks` bails on the `/proc/<pid>/task` read. Nothing corrects it afterwards, and `upgradeReconciledStart` has no test coverage.

`info.preDaemon = identityOK` leaves a refused container at the sentinel — the stricter outcome. If the upgrade should still apply after a transient failure, re-run `verifyContainerTask` inside `upgradeReconciledStart` and drop the comment.

cmd/container_attribution.go
```diff
@@ -0,0 +1,349 @@
...
+			if !aok {
+				continue
+			}
+			addr = addr.Unmap()
+			if !prefix.Contains(addr) {
+				continue
+			}
+			ones, _ := ipn.Mask.Size()
+			own, perr := addr.Prefix(ones)
+			if perr == nil && own == prefix {
```
jordanjennings: The exemption compares only the interface address's own masked network against the claimed prefix — no device name, no device type, no route lookup — so it can't distinguish "the subnet's own bridge" from any other interface whose network happens to equal the subnet. With `tun0 = 10.8.0.5/16`, `docker network create --subnet 10.8.0.0/16` computes `own == prefix` and is exempted, carving the VPN's range out of the hook. That contradicts what `bpf/originbpf.c:169-171` and design.md now advertise ("refused when the subnet contains a real host interface address").

Two qualifiers, so this is weighted correctly:

- The equivalent LAN case is *not* a bypass. TC attaches via `FindPodInterface()`, which prefers eth0, so eth0-bound traffic still meets TC's verdict. Only destinations routing over an interface TC isn't attached to — a tunnel, a second NIC — go unadjudicated. On a stock hosted runner (single eth0, no tunnels) this isn't reachable at all; it needs a self-hosted runner with a VPN or second NIC.
- Docker does not stop it. In moby, the host-derived reserved list is consulted only when `req.Pool == ""`, so an explicit `--subnet` skips it; `allocateSubnet`'s overlap check iterates only Docker's own pools, and `InferReservedNetworks`' doc comment names static allocation as the documented escape hatch. (Whether the destination stays reachable afterwards depends on which of the two competing link-scope routes the kernel prefers — if the bridge route wins it degrades to self-DoS. I haven't established that either way.)

Gating the exemption on the interface genuinely being a bridge — stat `/sys/class/net/<name>/bridge` — preserves `docker0`/`br-*` and rejects tun and physical devices, without plumbing the network name through.

Worth saying that this fix alone doesn't close the class: the check never consults routes, so a split-tunnel aggregate with no host address inside passes without needing the exemption; and with no delete path and no re-validation, claiming a prefix before a VPN comes up carves it permanently.

cmd/container_attribution.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// hostAddressInside reports a non-loopback host interface address contained
+// in prefix whose own network is not exactly prefix — the signature of a
+// carve-out claiming address space the host genuinely lives in. The
+// bridge's own gateway address matches prefix exactly and passes.
+func hostAddressInside(prefix netip.Prefix) string {
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return "" // can't validate: don't block the carve-out on it
```
jordanjennings: This fails open: an enumeration error returns "no conflict", so the carve-out is written without the locality check ever having run. Same on the `ifc.Addrs()` error a few lines down. Neither path logs — only the refusal branches do — so a validation that silently didn't happen is indistinguishable from one that passed.

Given the write is permanent (no delete path) and the check is one-shot, failing closed here costs a carve-out that container-driven discovery would retry on the next container anyway, which seems the better trade. At minimum log it.

pkg/containers/reconcile.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+		t.missingCounts[id]++
+		if t.missingCounts[id] >= 2 {
+			stale = append(stale, id)
+			delete(t.missingCounts, id)
+		}
+	}
+	tracked := len(t.containers)
+	t.mu.Unlock()
+	if len(list) == 0 && tracked > 0 {
```
jordanjennings: The empty-list guard returns after the counting loop has already run, so the degenerate pass still spends one of the two absences the new two-pass sweep exists to require.

Sequence: pass N returns empty — every tracked container's `missingCounts` goes to 1, then this guard returns and suppresses the sweep. Pass N+1 returns a partial-but-nonempty list omitting live container X — X's count reaches 2 and it is swept on its first *credible* absence. That's the settling-daemon case the change was written to prevent, and per the comment above the sweep is permanent (live-restore containers never re-emit a start event).

The comment at `:126-127` describes the guard as "the degenerate case of the same distrust", but it isn't — the degenerate case still consumes an absence. Checking `len(list) == 0` before the counting loop rather than after restores the intent.

pkg/containers/start_exec.go
```diff
@@ -0,0 +1,382 @@
...
+		}
+	}
+}
+
+func (t *Tracker) remove(id string) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	// Unconditional: exec ids may have been seen for containers the start
+	// handler never registered (inspect raced the container's death).
+	delete(t.seenExecs, id)
```
jordanjennings, line 376: `seenExecs` is deleted unconditionally here but `missingCounts` never is, which costs both ways.

Growth: an entry is created only for ids in `t.containers` and cleared only when the container reappears in a list or its count reaches 2. A container absent for one pass and then removed by `die`/`destroy` leaves an entry nothing ever collects — the counting loop iterates `t.containers`, which it is no longer in. Same unbounded-growth shape the per-container `seenExecs` keying and its unconditional delete were introduced to avoid.

Correctness: container ids survive `docker restart` and `stop`+`start`, so a re-adopted id carrying a stale count of 1 is swept on its first absence, defeating the two-pass guard for exactly that container.

`delete(t.missingCounts, id)` alongside the `seenExecs` delete, before the `info == nil` return, for the same reason that one is unconditional.

pkg/containers/network.go
```diff
@@ -0,0 +1,256 @@
...
+		nw, err := t.client.inspectNetwork(ictx, netName)
+		cancel()
+		if err != nil {
+			t.logger.Warn("Cannot inspect network for subnet carve-out; subnet stays adjudicated for now",
+				"network", netName, "subnet", key, "error", err)
+			return // unresolved: retried by the next container, not cached
+		}
+		isBridge = nw.Driver == "bridge"
+		t.mu.Lock()
+		t.netDrivers[netName] = isBridge
```
jordanjennings: The cache is keyed by network *name* and lives for the process. `handleNetworkCreate` refreshes it on create, but nothing evicts on destroy and `reconcile` has no network pass, so a name can outlive the network it described.

After `docker network rm foo` (bridge) → `docker network create -d macvlan foo`, if the create event is missed — a dockerd restart empties the buffer `since` replays from — a container on macvlan `foo` reads the stale `true` and its physical-network subnet is carved out of the hook. That's precisely what the driver gate exists to prevent, and `bpf/originbpf.c`'s comment leans on userspace getting this right ("Userspace writes only bridge-driver network subnets").

Partly mitigated by `hostAddressInside`, but only for macvlan ranges the host itself has an address in. Handling `network destroy` in the dispatch (delete the name from `netDrivers`) covers the common path; clearing `netDrivers` at the top of each `reconcile` also covers the missed-event case.

pkg/origin/origin.go
```diff
@@ -0,0 +1,732 @@
...
+func (o *Observer) lookup(key flowKey, srcPort uint16) []Record {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	entries := o.flows[key]
+	if len(entries) == 0 {
+		return nil
+	}
+	if srcPort != 0 {
+		for _, r := range entries {
+			if r.SrcPort == srcPort {
```
jordanjennings, line 747: The exact source-port match returns the newest matching record alone, and the doc comment above states the assumption it rests on — "MASQUERADE keeps the source port unless it collides" — without handling the collision it names.

When two containers in separate netns pick the same ephemeral source port to the same `dst:port` within the store window, both records live under one key with the same `SrcPort`. MASQUERADE preserves one and rewrites the other, so the TC event's post-NAT source port matches both records but identifies only one flow. The caller then sees a single candidate, takes the resolved branch rather than the ambiguous-containers branch, and stamps a container id and step ordinal as fact — the one case the comment says should degrade to the stricter tier ("disagreement must degrade… that policy belongs to the caller").

Narrow, and attribution-only. Returning all records whose `SrcPort` matches, rather than the first, hands the ambiguity to the caller that already knows how to file it.

pkg/origin/origin.go
```diff
@@ -0,0 +1,732 @@
...
+func (o *Observer) divertOverflow(rec Record, atClose bool) {
+	if rec.Verdict == VerdictBlock {
+		select {
+		case o.degradedCh <- rec:
+			return
+		default:
+		}
+	}
+	if atClose {
+		o.verdictsAbandoned.Add(1)
```
jordanjennings, line 689: A `VerdictBlock` that fails the degraded send falls through to this branch, so a genuinely lost enforce-mode audit record is counted under a label that denies the possibility. `:223` calls it "would-block reports discarded by Close's backlog policy" and `:355-360` says "not evidence of runtime saturation" — an operator reading the shutdown line has no signal that denials went unrecorded.

Reachable: `verdictCh` is 4096 deep and `degradedCh` 1024, and at Close `reportLoop` drains at memcpy speed while `degradedLoop` pays an audit write per record, so a backlog above ~1024 Blocks fills the lane and every subsequent one is counted as abandoned. The `Warn` at `:673` is in the non-`atClose` branch only, so the loss is otherwise silent.

Relatedly, `Record.Degraded`'s doc at `:126-132` says the lane exists so an enforce-mode drop "ALWAYS produces an audit record", which `degradedQueueDepth`'s own comment at `:173-179` contradicts ("Only when BOTH lanes are saturated is a Block's report finally dropped").

Counting a Block that failed the degraded send into `verdictsDropped` regardless of `atClose` — or a fourth counter — keeps "abandoned" meaning what it documents, and softening the `ALWAYS` reconciles the two comments.

pkg/dns/server.go
```diff
@@ -387,19 +534,37 @@ func (s *Server) handleDNSQuery(w dns.ResponseWriter, r *dns.Msg) {
...
+				// --docker-dns-interception is on, but without container
+				// attribution nothing could ever substantiate the claim —
+				// stamping it would misfile host clients of the bridge
+				// address into the container tier. With tracking wired, the
+				// classification stays per-listener (a lookup MISS keeps
+				// container origin: unknown-but-bridge is the unattributed
+				// container tier, and the sockdiag path must never run for
+				// a socket it cannot see).
+				containerLookup, containerWired := s.containerLookup.Load().(ContainerLookup)
+				if s.isContainerListener(w) && containerWired {
```
jordanjennings: Grouping four small ones here to keep the thread count down.

**This line.** Adding `&& containerWired` sends bridge-listener queries down the `else if` host sockdiag path whenever the container lookup isn't installed yet — a real window, since `SetContainerLookup` only lands in `startUserspace` after the dockerd restart (`cmd/start.go:842`), while the bridge listener exists from startup. That path isn't inert for a container address: `sockdiag.go:149-150` falls back to `if wildCount == 1 { return wildCookie, true, nil }`, so a single wildcard-bound host socket sharing the container's ephemeral source port yields an unrelated host process's ordinal. Low probability (wildcard-bound sockets are usually listeners on well-known ports), but it's a wrong non-zero ordinal rather than a miss, and in that same window the query also loses the `ContainerOrigin` flag the previous code set. Keeping `ContainerOrigin` on the listener test and gating only the lookup on `containerWired` would preserve the tier without reaching for sockdiag.

**`pkg/containers/network.go:223`** — `DiscoverBridgeSubnets` builds a private `dockerClient` whose `http.Transport` is never closed and becomes unreachable on return, holding an idle unix-socket connection for the process lifetime. `defer c.http.CloseIdleConnections()`.

**`cmd/start.go:1201`** — this comment says the DNS-infra auto-allows "are gated on the conditions that create the listeners", but only the bridge entry is; `127.0.0.1` and the `--dns-upstream` host are unconditional, as the block at `:453-469` says in as many words. Two comments added in the same change describing the same code incompatibly.

**`pkg/config/config.go`** — `cidrDenyCoversLocked` only considers all-ports denies, so a port-limited operator deny (say `deny 10.0.0.0/8` port 53) falls through to `cidrCoveredLocked`, which skips the action check when `port != nil` and reports the target as covered. Suppressing the auto-allow is the right outcome, but it logs "Allow rule already exists" — false, and on the exact line someone greps when DNS stops resolving.

jordanjennings (approved):
All three blockers from the previous review are fixed and CI is green on `1c623c6`. The comments below are non-blocking.

One cross-cutting note: the new safety machinery is untested. `hostAddressInside`, the width cap, `upgradeReconciledStart`, and the local-nets carve-out's isolation property have no coverage — and the isolation test that does exist cannot fail (see the comment on `originbpf_test.go`).

bpf/originbpf_test.go
matthewdevenny: Right — two independently loaded collections made the assertion vacuous. Fixed in 4a8d35f: the test now loads origin via `loadOriginObjectsSharing` (the production `MapReplacements` wiring) and seeds policy through the shared maps, so moving `map_local_nets` into the shared set — the regression it advertises catching — makes TC consult the carve-out and the final require fires.

Related, from the structural review that followed yours: the entire validation policy now lives ON `AllowLocalNetwork` itself (`pkg/origin/localnets.go`), so the property this test pins and the policy that preserves it are one package, and no future caller can reach the map without passing the check.

pkg/containers/start_exec.go
matthewdevenny: Took the strict option in 4a8d35f: `info.preDaemon = identityOK`, so a refused container stays at the sentinel and the upgrade path's "identity was verified at adoption" premise is true again. Coverage is end-to-end, not just mechanics: `TestIdentityMismatchBlocksReconcileUpgrade` drives reconcile-adoption-with-refused-identity → replayed start event through the fake daemon and proves no tag ever reaches pid 4242 (plus `TestUpgradeReconciledStart` for the upgrade mechanics themselves — ordinal from event time, exec re-tag survival, pre-step no-op).

cmd/container_attribution.go
matthewdevenny: Fixed in 4a8d35f exactly as suggested: the `own == prefix` exemption additionally requires the device to genuinely be a linux bridge (`/sys/class/net/<name>/bridge`), so `tun0 = 10.8.0.5/16` claiming `10.8.0.0/16` is refused — `TestValidateLocalSubnet` pins that exact case alongside the width-cap and VPC-containment ones. The residual you flagged (addresses not routes; a prefix claimed before the VPN comes up is carved permanently) is documented at the validator rather than papered over.

The check also moved out of cmd and onto the map write itself (`pkg/origin/localnets.go`) per the follow-up structural review, so the choke point is the API, not a caller.

cmd/container_attribution.go
matthewdevenny: Fails closed now (4a8d35f): interface-enumeration errors — including per-interface `Addrs()` failures — return an error from `AllowLocalNetwork`, which the tracker treats as transient and retries on the network's next container. Policy refusals are a distinct sentinel (`origin.ErrLocalNetworkRefused`, logged at the write, stamped seen and never retried), so refused-vs-failed is now a package contract instead of a convention in one caller.

pkg/containers/reconcile.go
matthewdevenny: Fixed in 4a8d35f: the empty-list judgment moved before the counting loop, so the degenerate pass no longer spends one of the two absences and the next partial list cannot sweep a live container on its first credible absence. The comment was rewritten to say what the guard actually does (and a duplicated paragraph from an earlier edit was collapsed while there).

pkg/containers/start_exec.go
matthewdevenny: `delete(t.missingCounts, id)` is unconditional in `remove` now (4a8d35f), alongside the `seenExecs` delete and for the same reasons — both the leak and the sharper failure you called out (docker ids survive stop+start, so a stale count of 1 would defeat the two-pass guard for exactly the re-adopted container). `TestRemoveClearsMissingCount` covers both the registered and the never-registered id.

pkg/containers/network.go
matthewdevenny: Both suggestions taken in 4a8d35f: `network destroy` events evict the name from the cache, and reconcile clears `netDrivers` wholesale at the top of each pass — reconcile being the recovery point for exactly the missed-event window you described (dockerd restart empties the buffer `since` replays from). The cache also gained a single owner: `Tracker.carvePrefix` is now the one apply path every discovery source funnels through, so the seen/driver/stamp pipeline exists once.

pkg/origin/origin.go
matthewdevenny: Fixed in 4a8d35f: the exact-port pass returns EVERY record matching the source port. A single match keeps its authoritative fast path; a collision hands all candidates to `resolveJoin`, which already knows how to degrade the disagreement (container-only when they agree on container, ambiguous tiers otherwise, PID dropped on PID disagreement). `TestLookupExactSourcePortCollisionReturnsAllMatches` pins it, and the doc comment now names the collision as handled rather than assumed away.

pkg/origin/origin.go
matthewdevenny: Fixed in 4a8d35f: a Block that misses both lanes counts as `verdict_reports_dropped` with its own warning regardless of `atClose` — "abandoned" now only ever means would-block shutdown bookkeeping, so the shutdown line can no longer hide unrecorded denials. The `ALWAYS produces an audit record` doc was softened to match `degradedQueueDepth`'s own caveat: the record survives unless BOTH lanes are saturated, and that loss is counted where saturation is counted.

pkg/dns/server.go
matthewdevenny: All four in 4a8d35f:

**This line** — went one step past the suggestion, per the structural review that followed: attribution is now a constructor-time property of the listen address (`listenerAttribution`: `attributeContainerIP` vs `attributeHostSockdiag`, set by `AddContainerListenAddr`/`AddListenAddr`), and `handleDNSQuery` switches on the listener's mode. Bridge queries never reach sockdiag in any window — your wildcard-fallback wrong-ordinal scenario is structurally unreachable — and the lookup alone is gated on being wired. The accepted residual (interception without attribution files bridge queries, including host clients of the bridge address, as unattributed container traffic) is pinned by a test named for it: `TestHandleDNSQuery_InterceptionWithoutAttributionFilesAsUnattributedContainer`.

**network.go** — `defer c.http.CloseIdleConnections()` added.

**start.go comment** — fixed to say the DNS-infra allows run unconditionally.

**config.go** — `cidrDenyCoversLocked` extended to port-limited denies (port set + protocol overlap, equal-or-broader prefix), so your `deny :53` example now suppresses with the honest "deny rule covers this network" log instead of the misleading dedup line. Five-case table test added — which promptly caught me once: a deny on a *different protocol* correctly does not veto, and my first table expectation said otherwise.

matthewdevenny:
All ten inline comments plus the cross-cutting test-coverage note are addressed in `4a8d35f` — details in the thread replies. On the coverage note specifically: the isolation test now genuinely shares the verdict maps (it could not fail before, as called out), and the new safety machinery each got tests — `TestValidateLocalSubnet` (width cap, VPC containment, bridge self-exemption, the tun0 exact-match refusal), `TestIdentityMismatchBlocksReconcileUpgrade` (end-to-end through the fake daemon), the source-port-collision lookup test, and a five-case deny-veto table for `cidrDenyCoversLocked`.

A structural review round landed in the same commit and moved several of these fixes to better owners: the carve-out policy now lives ON `origin.AllowLocalNetwork` (`pkg/origin/localnets.go`) with a refused-vs-failed error contract (`ErrLocalNetworkRefused`), all three discovery sources funnel through one `Tracker.carvePrefix` apply path, DNS listener attribution is a constructor-time mode rather than a branch, and `pkg/containers` was decomposed (tracker/reconcile/start_exec/network/enricher, all well under 1k lines).

PR #111, opened 2026-08-20
Add conntrack flush

matthewdevenny:
Closes the DNS-redirect bypass behind the unattributed CI block of `150.171.109.74` (2026-08-20) — an IP the DNS proxy provably never served. The nat table is evaluated only on a flow's first packet; the verdict is stamped into its conntrack entry and replayed thereafter. DNS flow state predating the DNAT rules therefore keeps steering queries straight to the allow-listed upstream resolver, silently bypassing the proxy's rule matching and IP allowlisting — and a bypassed answer can land in systemd-resolved's cache and serve clients long after the flow itself is gone.

## Change

- **`pkg/network/conntrack.go`** (new): `FlushDNSConntrack` deletes every IPv4 conntrack entry with original-direction dport 53 (TCP+UDP). ctnetlink over the existing `mdlayher/netlink` dep: dump, filter at parse time, per-entry ACKed delete. Dump-then-delete is the only mechanism the kernel offers — the flush path rejects `CTA_FILTER` with EOPNOTSUPP (verified on 6.8), which is also why `conntrack -D --dport` iterates in userspace.
- **`SetupDNSRedirect`** flushes right after installing the DNAT — before the resolved-cache flush in start.go, so no bypassed answer can arrive after the cache is purged. **`TeardownDNSRedirect`** flushes so flows still DNAT'd to the dead proxy don't blackhole client DNS until entry expiry.
- Scoped to port 53 (a full flush would churn Docker MASQUERADE state) and best-effort with honest failure behavior: bounded deadlines, a 3-attempt dump retry on a fresh socket, immediate bail on permanent errors (EPERM / EOPNOTSUPP), and partial flushes reported as `flushed N of M` in the call-site Warn. A failed flush never blocks startup.
- Runs only when the redirect does (`--dns-redirect-iptables`, implied by both CI presets); default standalone runs are untouched. README's manual-wiring option gains the matching caveat + `conntrack -D` one-liner; flush mechanics, collateral analysis, and measurements are documented in design.md.

## Verified

- Unit tests: marshal/parse round-trip, incomplete and malformed payloads, scope filter.
- `TestFlushDNSConntrack_Live` runs kernel-backed **in CI** via the sudo test pass (`./pkg/network/` added to `make test-ci`), so the ctnetlink wire format is pinned on every run: an ESTABLISHED DNS-over-TCP stream and a UDP flow to :53 both die; TCP and UDP controls on :5353 survive.
- e2e Setup/Teardown harness on kernel 6.8 found 33 genuinely stale DNS flows on the dev VM; a 5k-entry dump with the default rcvbuf completes in ~5ms; full linux suite green.

The Go 1.26 / dependency bump originally on this branch landed separately as #112; this branch builds and passes against both the old and new dependency pins.


mtmk:
Read through this and cross-compiled the branch for linux (vet and build clean). The mechanism looks right to me: the attribute layout matches the uapi headers, ports are raw big-endian without NLA_F_NET_BYTEORDER (correct), the decoder masks NLA_F_NESTED off Type() so kernel messages parse, and delete-by-original-tuple is the right way to address a DNAT'd entry. `errors.Is(err, unix.ENOENT)` does unwrap, since OpError.Err carries a unix.Errno. I also checked that nothing in the tree matches on `-m conntrack --ctstate`, so deleting an entry can't get a reply dropped by a state rule, and that the proxy is listening before SetupDNSRedirect runs, so flushed flows re-DNAT onto something alive.

Things I'd like addressed:

1. The live test never runs in CI. `make test-ci` runs its sudo pass over `./bpf/` and `./pkg/tc/` only, so pkg/network runs unprivileged and TestFlushDNSConntrack_Live always skips. The ctnetlink wire format is exactly the thing that breaks silently on a kernel or netlink bump, and right now the only thing verifying it is a manual Lima run. Adding ./pkg/network/ to the sudo pass would fix that; the other tests in the package use stubbed binaries and should be fine as root. Worth calling out that the live test flushes the runner's own dport-53 conntrack, which is harmless by the same argument the change makes, but should be a deliberate choice rather than a surprise.

2. bpf/originbpf_bpfel.o is regenerated with no .c change, and the other five objects aren't. verify-bpf-generated-code only triggers on the .c/.h paths, so nothing here revalidates it and it landed from a local clang rather than the pinned CI environment. Either drop the object from the commit or add go.mod to that workflow's path filter so a toolchain bump re-verifies generated artifacts.

3. The dependency and toolchain commit (Go 1.25 -> 1.26, nine direct deps, yaml.v3 -> go.yaml.in/yaml/v3, bpf regen) is unrelated to the bypass fix. It roughly doubles the surface to review and makes this harder to backport if the fix needs to go out on its own. I'd split it. DEPENDENCIES.md does line up with the new indirect block, so no issue there.

4. "an in-flight transaction costs at most one retry" holds for UDP but not for an established DNAT'd DNS-over-TCP stream: once its entry is gone, later packets miss the NAT verdict and the stream resets. Breaking it is the right call at both call sites, but that sentence is the justification for the blast radius, so it should say reset rather than retry.

5. isDNS() is a bit wider than the rules it is repairing. The DNAT rules carry `! -d 127.0.0.0/8`, so stub traffic to 127.0.0.53:53 has no verdict to fix, yet it gets deleted anyway. Harmless, but it undercuts the "collateral-free" claim right above it. Either exclude loopback destinations or soften the comment.

6. Permanent failures get retried as transient. A non-root run (EPERM) or a kernel without nf_conntrack_netlink burns all three dump attempts back to back. Cheap, but bailing on EPERM/EOPNOTSUPP would be more honest.

7. One 2s deadline covers every delete, not each. On a table with many DNS entries the budget can expire mid-loop, after which the remaining deletes fail immediately and the flush finishes partially, with the returned error carrying no sense of how far it got. Stopping on a deadline error and putting attempted/deleted counts in the warn would make a partial flush visible in the logs.

Test coverage itself is good: round-trip against the marshaller, incomplete and malformed payloads, the scope table, and a live test that asserts both the kill and the survivor.

mtmk (approved):
LGTM

pkg/network/conntrack.go
```diff
@@ -0,0 +1,319 @@
...
+// FlushDNSConntrack deletes every IPv4 conntrack entry whose original-
+// direction destination port is 53 (TCP or UDP). nat is evaluated only on a
+// flow's first packet and the verdict replayed from its conntrack entry, so
+// DNS flow state predating the redirect bypasses the proxy — and state
+// surviving teardown keeps DNAT'ing to the dead proxy — until deleted.
+// Scoped to port 53 (a full flush would churn unrelated NAT state) and
+// best-effort at both call sites: the iptables rules stand either way, but
+// an actively-used stale flow can bypass indefinitely, hence the Warn.
+// Mechanics, collateral analysis, and measurements: design.md.
+func FlushDNSConntrack(logger *slog.Logger) error {
```
jordanjennings, line 349: The delete only repairs a flow whose next packet comes from the client. If the next packet comes from upstream (the reply to a query in flight at flush time, or a server-side ACK on a DNS-over-TCP stream), conntrack creates a new entry with upstream as ORIGINAL; nat PREROUTING/INPUT have no rules here so `nf_nat_alloc_null_binding` stamps it, and every later client packet on that socket is `IP_CT_ESTABLISHED_REPLY`, which `nf_nat_inet_fn` never runs the nat OUTPUT chain for. That socket bypasses the DNAT for as long as it keeps the entry refreshed, and the entry's original tuple now has dport = the client's ephemeral port, so `isDNS()` can never select it for a later flush.

For per-query sockets that is a one-off. For a client that reuses one UDP socket (c-ares, i.e. Node's `dns.resolve*`, keeps one ephemeral port per server indefinitely by default, and Node is one of the callers this redirect exists for) or an established DNS-over-TCP stream, it is a persistent bypass, and the in-flight ordering is the likely one on a busy runner. The design.md text this PR adds ("an in-flight UDP transaction costs one retry", "an established DNAT'd DNS-over-TCP stream is reset") only holds when the client speaks first.

One way to close it without changing the flush: refuse to confirm reverse-direction DNS entries, e.g. `iptables -A INPUT -p udp --sport 53 -m conntrack --ctstate NEW -j DROP` (plus the TCP equivalent) installed alongside the DNAT rules. A dropped packet's unconfirmed entry is destroyed, so the client's next packet is NEW in the forward direction and takes the DNAT; the cost is exactly the one retry / one reset design.md already budgets.

pkg/network/conntrack.go
```diff
@@ -0,0 +1,319 @@
...
+		return fmt.Errorf("marshal conntrack tuple: %w", err)
+	}
+	_, err = conn.Execute(netlink.Message{
+		Header: netlink.Header{
+			Type:  ctMsgType(ipctnlMsgCtDelete),
+			Flags: netlink.Request | netlink.Acknowledge,
+		},
+		Data: append(nfgenmsgV4(), attrs...),
+	})
+	if err != nil && !errors.Is(err, unix.ENOENT) {
```
jordanjennings: ENOENT is folded into success and the caller counts it in `deleted`, so a systematic ENOENT reports as a completed flush. One concrete route: entries in a non-default conntrack zone (`-j CT --zone`, OVS) are in the dump, but `marshalTupleOrig` never echoes `CTA_ZONE`, and `ctnetlink_del_conntrack` defaults an absent zone to 0 and returns ENOENT. Counting ENOENT separately from real deletes keeps the race case quiet while making the systematic case visible; echoing `CTA_ZONE` from the dump fixes the zone case itself. Non-blocking.

pkg/network/conntrack.go
```diff
@@ -0,0 +1,319 @@
...
+				return lastErr
+			}
+			continue
+		}
+		deleted := 0
+		lastErr = nil
+		for _, t := range tuples {
+			if err := deleteTuple(conn, t); err != nil {
+				lastErr = err
+				// The attempt's deadline covers the whole delete loop;
```
jordanjennings: The previous revision re-armed a fresh `ctFlushTimeout` before the deletes; now the dump and the whole delete loop share the one 2s deadline from `dialCt`, and a delete-phase error returns at line 311 without using the remaining attempts (the retry loop only re-runs the dump). A slow dump leaves the deletes whatever is left, and the timeout they then hit is the case a fresh socket would have fixed. Re-arming after the dump restores that. Non-blocking.

pkg/network/conntrack.go
```diff
@@ -0,0 +1,319 @@
...
+
+// dumpTuples dumps the IPv4 conntrack table over conn, keeping the tuples
+// keep() accepts (nil keeps all). Best-effort by design: ctnetlink dumps
+// carry no consistency signal (the conntrack dump path never stamps
+// NLM_F_DUMP_INTR), so an entry that moves during the walk can be missed —
+// the flush makes no completeness claim. Unparseable entries are skipped,
+// not fatal: an entry we cannot address for deletion is one we could not
+// act on anyway.
+func dumpTuples(conn *netlink.Conn, keep func(ctTuple) bool) ([]ctTuple, error) {
+	msgs, err := conn.Execute(netlink.Message{
```
jordanjennings, line 238: `Execute` materializes the whole dump before `keep` runs (each `Message.Data` pins its recv buffer), so a large table is held in full to find a handful of dport-53 entries. netlink v1.11.2's `Conn.ReceiveIter` lets the filter run per message. Non-blocking.

pkg/network/dns_redirect.go
```diff
@@ -145,5 +152,12 @@ func TeardownDNSRedirect(logger *slog.Logger) error {
 	if lastErr == nil {
 		logger.Info("DNS redirect iptables rules removed")
 	}
+
+	// Purge flows still DNAT'd to the now-dead proxy so client DNS recovers
+	// immediately instead of at conntrack expiry (see FlushDNSConntrack).
+	if err := FlushDNSConntrack(logger); err != nil {
```
jordanjennings, line 193: This runs on every shutdown, including when `SetupDNSRedirect` failed and nothing was installed (`cmd/start.go` registers the teardown before the mutation and treats setup failure as a warning), so cargowall deletes the host's dport-53 flow state on a run where it never touched DNS, and warns that client DNS may stall for a redirect that was never there. It also runs when the `-D` calls failed and the DNAT rules are still in place, where the flush only recreates entries pointed at the dead proxy. Gating on the rules actually having been removed (`lastErr == nil`, or a flag set by setup) covers both.

Pre-existing rather than from this PR: the loop above deletes in install order, so the mark RETURN rules go before the DNAT rules while the proxy is still up (`dnsCancel` is deferred earlier in `startCargoWall`, so it unwinds after this). For the span of two `iptables -D` execs the proxy's own upstream queries match the DNAT and are sent back to itself. Iterating `dnsRedirectRules` in reverse for `-D` closes it.

pkg/network/conntrack_test.go
```diff
@@ -0,0 +1,270 @@
...
+	// remove the conntrack entry within the test window. TCP uses real
+	// listeners so the :53 flow is ESTABLISHED, not a refused SYN.
+	udpDNSPort := sendUDP(t, "127.0.0.2:53")
+	udpOtherPort := sendUDP(t, "127.0.0.2:5353")
+	listenTCP(t, "127.0.0.2:53")
+	listenTCP(t, "127.0.0.2:5353")
+	tcpDNSPort := dialTCP(t, "127.0.0.2:53")
+	tcpOtherPort := dialTCP(t, "127.0.0.2:5353")
+
+	if !hasTuple(t, unix.IPPROTO_UDP, udpDNSPort, 53) {
```
jordanjennings: The skip condition goes through the parser under test: `hasTuple` → `dumpTuples` → `parseConntrackTupleOrig`. A wire-format regression that makes the dump yield zero tuples (a symmetric marshal/parse bug passes the round-trip test) reads as "conntrack not tracking loopback" and skips green, which is the failure the new sudo lane was added to catch. Skip only when the dump is empty; if the table has entries but not the planted flow, fail. Non-blocking.

Makefile
```diff
@@ -73,7 +73,7 @@ test-ci:
 	@printf "${GREEN}Running CI tests...${RESET}\n"
 	go run gotest.tools/gotestsum@latest --junitfile test-results.xml --format testdox -- ./...
 	@printf "${GREEN}Running BPF tests with sudo...${RESET}\n"
-	sudo go run gotest.tools/gotestsum@latest --junitfile test-results-bpf.xml --format testdox -- -count=1 ./bpf/ ./pkg/tc/
+	sudo go run gotest.tools/gotestsum@latest --junitfile test-results-bpf.xml --format testdox -- -count=1 ./bpf/ ./pkg/tc/ ./pkg/network/
```
jordanjennings, line 76: `test-bpf` (line 70) is the local counterpart of this lane and still runs only `./bpf/ ./pkg/tc/`.

jordanjennings (changes requested):
Requesting changes for the reverse-direction pickup in `FlushDNSConntrack`; everything else inline is non-blocking.

pkg/network/conntrack.go
matthewdevenny: Fixed in a71bbbf with the guard you proposed: `dnsRedirectRules` now carries per-rule chains and installs `INPUT -p udp/tcp --sport 53 -m conntrack --ctstate NEW -j DROP` alongside the DNAT — in place before the flush runs, so a reply in flight across it can't confirm a reversed entry. design.md documents the mechanism (null binding, reply-direction packets never traversing nat OUTPUT, ephemeral-port orig tuple invisible to later flushes), and the README manual-wiring bullet now tells option-3 users to pair their DNAT with the same guard. The e2e harness asserts the guard installs with setup and is gone after teardown. You were also right that the design.md cost claims only held when the client speaks first — that text now says the guard makes the retry/reset budget direction-independent.

pkg/network/conntrack.go
matthewdevenny: Both taken in a71bbbf: `ctTuple` carries the zone, parsed from top-level `CTA_ZONE` in the dump and echoed on delete (the round-trip test covers a zoned tuple), and ENOENT acks are counted separately — the success log reads `count=N already_gone=M`, so the benign dump→delete race stays quiet while `already_gone` dominating `count` flags a systematic addressing failure. A delete-phase error additionally reports `flushed N of M`.

pkg/network/conntrack.go
matthewdevenny: Restored in a71bbbf, and the structure around it changed in a way that covers the rest of this thread too: retry is now whole-attempt — `flushAttempt` dials, dumps, and deletes on one socket (`defer Close`), and `FlushDNSConntrack` retries the complete attempt on a fresh socket, so a mid-delete deadline gets the remaining attempts instead of returning early. Each attempt re-arms `ctFlushTimeout` between dump and deletes.

pkg/network/conntrack.go
matthewdevenny: Keeping `Execute` here, deliberately: the materialized dump is bounded and transient (~1.5MB for a 5k-entry table, held for single-digit milliseconds on the startup path), and `Execute` keeps sequence validation and multipart handling inside the library — a hand-rolled receive loop is exactly the machinery earlier review rounds pushed this file to shed. Happy to revisit with `ReceiveIter` if a deployment shows table sizes where the difference is real.

pkg/network/dns_redirect.go
matthewdevenny: Both halves taken in a71bbbf: teardown deletes in reverse install order, so the mark RETURN exemptions outlive the DNAT rules they exempt — closing the two-exec self-DNAT window — and the flush is gated on at least one `-D` having succeeded rather than all: a partial setup that left DNAT installed still gets its flush, while a run that never installed anything (`removed == 0`) leaves the host's flow state alone. The e2e harness covers teardown-without-setup: per-rule warns, no flush, error returned.

pkg/network/conntrack_test.go
matthewdevenny: Taken in a71bbbf: the skip now triggers only on an empty dump (conntrack not active); a non-empty table missing any of the four planted flows fails with "planted flow missing from a non-empty dump — parser or marshal regression" — exactly the symmetric-bug case you described.

Makefile
matthewdevenny: Taken in a71bbbf — `./pkg/network/` added to `test-bpf` as well, so the local root lane matches the CI one.

pkg/network/dns_redirect.go
```diff
@@ -33,30 +33,54 @@ const DNSProxyFWMark = 0xCA12
...
 	// Redirect all other outbound DNS to the local proxy
-	{"-t", "nat", "-p", "udp", "--dport", "53", "!", "-d", "127.0.0.0/8", "-j", "DNAT", "--to-destination", "127.0.0.1:53"},
-	{"-t", "nat", "-p", "tcp", "--dport", "53", "!", "-d", "127.0.0.0/8", "-j", "DNAT", "--to-destination", "127.0.0.1:53"},
+	{"OUTPUT", []string{"-t", "nat", "-p", "udp", "--dport", "53", "!", "-d", "127.0.0.0/8", "-j", "DNAT", "--to-destination", "127.0.0.1:53"}},
+	{"OUTPUT", []string{"-t", "nat", "-p", "tcp", "--dport", "53", "!", "-d", "127.0.0.0/8", "-j", "DNAT", "--to-destination", "127.0.0.1:53"}},
+	// Reverse-direction guard: an inbound sport-53 packet must not CREATE a
+	// flow — a reply racing the conntrack flush would re-create it reversed,
+	// with a null NAT binding no later flush can see (design.md, "DNS
+	// redirect"). Replies on ESTABLISHED entries never match.
+	{"INPUT", []string{"-p", "udp", "--sport", "53", "-m", "conntrack", "--ctstate", "NEW", "-j", "DROP"}},
```
jordanjennings: The guard has no interface scope, so it also applies on `lo`. Loopback-destination entries (the 127.0.0.53 stub, or a resolv.conf pointed straight at 127.0.0.1) are still in the flush's scope, and until now deleting one mid-transaction was the no-op design.md describes: the reply recreated the entry in reverse, harmlessly, since nothing on `lo` is DNAT'd. With the guard installed ahead of the flush, that reply is now NEW on `lo` and dropped, so a process with a stub lookup in flight across the install flush loses the answer and sits out its resolver timeout (5s for glibc). Nothing on `lo` needs the guard: a reverse entry from a 127.x:53 reply has reply-tuple dst 127.x, which a client's next query to an external resolver never matches, so that query is NEW forward and takes the DNAT. `! -i lo` on both INPUT rules closes it and keeps the live test's loopback target valid; the alternative is excluding 127/8 destinations from `isDNS()`, which then needs a non-loopback target in the live test. Either way the design.md sentence "deleting them is a no-op that recreates identically" no longer holds as written. Non-blocking.

pkg/network/dns_redirect.go
```diff
@@ -33,30 +33,54 @@ const DNSProxyFWMark = 0xCA12
...
+	{"INPUT", []string{"-p", "tcp", "--sport", "53", "-m", "conntrack", "--ctstate", "NEW", "-j", "DROP"}},
```
jordanjennings: The three teardown assertions in `api-failure-e2e.yml` ("DNS DNAT rules leaked after stop", the startup-SIGTERM check, the forced-teardown check) grep only `nat -S OUTPUT` for `127.0.0.1:53`, so a leaked INPUT guard passes them. `iptables -S INPUT | grep -q -- '--sport 53'` alongside each covers the new mutation. Non-blocking.

jordanjennings (approved):
Approving; two non-blocking notes inline.

PR #112, opened 2026-08-20
Update Go to 1.26 and dependencies

matthewdevenny:
Split out of #111 per review — the toolchain/dependency bump is unrelated to the DNS-conntrack fix and lands separately so the fix stays small and backportable.

## Change

- `go 1.26.0` directive; **mdlayher/netlink 1.7.2 → 1.11.2** (fixes a panic and a false rejection on unaligned conntrack messages — relevant to #111's consumer), **cilium/ebpf 0.21 → 0.22**, plus kong, miekg/dns, testify, x/net, x/sys, genproto, protobuf.
- `bpf2go` regeneration for ebpf 0.22: adds object-name constant blocks to the generated loaders (deterministic Go text). The compiled `.o` objects are untouched — no `.c`/`.h` changed, and objects must only ever be rebuilt in the pinned CI environment: a local cross-arch clang emits different bytes for identical source, which the newly-widened path filter caught on this PR's first push. Mockery regen is import-order only.
- `DEPENDENCIES.md` indirect table synced: testify 1.12 drops spew/difflib and moves yaml to `go.yaml.in/yaml/v3`; netlink 1.11 drops `josharian/native`.
- `verify-bpf-generated-code` now also triggers on `go.mod`, so a toolchain or dependency bump re-verifies generated artifacts from the pinned CI environment instead of trusting local clang (review point on #111).

## Verified

Full linux suite + BPF root suite in Lima under Go 1.26.7 (sole failure is the VM's known pre-existing Truncated-IPv6 EINVAL quirk); `go mod verify` clean; staticcheck (rebuilt for go1.26) clean across the repo.


mtmk (approved):
LGTM

Issue #127, opened 2026-09-15
Search-expanded forms of allowed single-label names are REFUSED: c-ares clients never reach the bare name


matthewdevenny:
## Symptom

With a search list in `resolv.conf`, every stub resolver asks `<name>.<search>` before `<name>` when the name has fewer dots than `ndots`. Under enforce with query filtering, the gate REFUSES that expanded form unless a rule happens to cover it. glibc and Go treat REFUSED as "try the next form"; c-ares does not — a REFUSED on a multi-label attempt ends its search (`src/lib/ares_search.c`, `search_callback`: the issue #852 carve-out for REFUSED/SERVFAIL applies to single-label attempts only). So on such a host:

- a single-label allow rule `myhost` does **not** make `myhost` resolvable for c-ares clients — the bare name is never tried;
- audit reports a `dns_blocked` for `myhost.<search>`, a name the operator never asked about, for every lookup.

Affected: Node `dns.resolve*` (not `dns.lookup`), Python `grpcio`, curl and gRPC builds on c-ares.

Reproduced on the #126 branch before its synthetic-specific handling, Lima VM, enforce + filtering, `search lan`:

```
c-ares (pycares)  _gateway: error=6 (DNS server refused query)
glibc (getent)    _gateway: 192.168.5.2          ← same second
proxy log         7× "DNS query blocked (domain not allowed)" domain=_gateway.lan
```

#128 closes this for the synthetic names by answering `<synthetic>.<host search suffix>` NXDOMAIN locally — correct there because the bare name is what resolves. This issue is the general form, which needs the opposite treatment: for a real host the expanded form is what resolves.

## Why the gate does what it does

The gate judges the name as asked, after stripping the *configured* search domains — the operator's `search-domains` plus the Kubernetes defaults (`StripSearchDomains`). The host's own `resolv.conf` search list is never consulted, so `myhost.lan` is not `myhost` for matching purposes, and a denied query is REFUSED. REFUSED is the honest answer for a name policy denies; it is the wrong rcode for a form the client is merely iterating through on its way to a name policy allows.

## Proposal

[...]

PR #128, opened 2026-09-15
#126 answer systemd-resolved synthetic names in the DNS proxy

### LINKED ISSUE #126: Resolver synthetic records die at the stub DNAT: `_gateway` stops resolving, silently in audit mode
## Symptom

systemd-resolved synthesizes a handful of names that exist in no zone and are never sent upstream. The DNS redirect DNATs the resolved stub to our proxy (pkg/network/dns_redirect.go:75-76), and the proxy's upstream is deliberately *not* the stub — `detectDnsUpstream` classifies `127.0.0.53` as loopback (cargowall-action src/dns.ts:47) and falls back to `/run/systemd/resolve/resolv.conf`, the real upstream, mirroring `detectSystemdResolvedUpstreams` (cmd/start.go:1785). Net effect: every resolved synthetic record stops resolving for the length of the run.

Verified on the Lima box (Ubuntu, systemd-resolved, `search lan`):

```
$ resolvectl query _gateway
_gateway: 192.168.5.2        -- link: eth0
-- Data from: synthetic

$ resolvectl query _outbound        →  192.168.5.15   -- Data from: synthetic
$ resolvectl query _localdnsstub    →  127.0.0.53     -- Data from: synthetic

$ grep ^hosts /etc/nsswitch.conf
hosts:          files dns                     ← no libnss-resolve, so this is a real UDP/53 lookup

$ dig @1.1.1.1 _gateway
;; ->>HEADER<<- opcode: QUERY, status: NXDOMAIN
```

`localhost` is synthetic too but survives, because `files` comes first in nsswitch and `/etc/hosts` carries it. `_gateway`, `_outbound` and `_localdnsstub` exist nowhere else, so with the redirect installed they NXDOMAIN.

Both postures break, differently:

- **enforce** — no hostname rule can match (`isQueryAllowed`, pkg/dns/server.go:535-575), so the query is REFUSED and logged as `dns_blocked`.
- **audit** — the query is forwarded, the real upstream NXDOMAINs it, and **no event is recorded at all**. The name fails and the run report is clean.

The audit case is the bad one: the failure is invisible in exactly the mode operators use to decide whether enforce is safe.

## Where it bites

Found while answering a customer question about Blacksmith GitHub Actions runners. Their microVMs reach a host-side agent — sticky disks, job manager, log ingestion, metrics, cache manager, Bazel RE cache, VM metadata — and inject its address as `BLACKSMITH_AGENT_ADDR`. Confirmed from useblacksmith's own public CI logs (`useblacksmith/checkout` container job):

```
-e "BLACKSMITH_AGENT_ADDR=192.168.127.1"
-e "BLACKSMITH_BAZELRE_CACHE_GRPC=grpc://192.168.127.1:1055"
-e "BLACKSMITH_ACTIONS_RESULTS_URL=http://192.168.127.96:49193/"
```

`192.168.127.1` is fixed across every host we sampled (10 distinct machines, two regions) — it's the gvisor-tap-vsock gateway inside each microVM, not a per-tenant allocation. Ports vary per VM (1026-1080, plus 49193 and 64001). Some runner generations name that host `_gateway` rather than the literal IP, and that is the configuration this issue is about.

Failure signature for the operator: their build gets slower and loses cache hits, while the Blacksmith actions warn and silently degrade —

```
BLACKSMITH_AGENT_ADDR or BLACKSMITH_STICKY_DISK_GRPC_PORT is not set; the Blacksmith
agent is not reachable from this runner, falling back to actions/checkout behavior
```

— and CargoWall reports nothing, because no connection was ever attempted.

## Why the design is what it is

Neither half is wrong on its own. The stub DNAT exists for a real reason (dns_redirect.go:70-74): a stub-following client otherwise reaches resolved, which re-queries upstream from its **own** socket, laundering the querying process away from the sockdiag join and serving warm cache hits the proxy never sees. And skipping `127.0.0.53` as an upstream is what stops the proxy from looping back into itself.

The consequence is that we take over the stub's role without taking over its full behaviour. resolved answers synthetics before it answers anything else; our proxy answers none of them.

## Proposal

**1. Answer resolved's synthetic names in the proxy.**

Handle them in `handleDNSQuery` before the filter gate and before the cache (pkg/dns/server.go:535), the same way the REFUSED path constructs a message locally (:569):

- `_gateway` → the default-route gateway addresses, ordered by metric
- `_outbound` → the source address chosen for outbound traffic
- `_localdnsstub` → `127.0.0.53`

Gate the whole thing on resolved actually running — probe `/run/systemd/resolve`, the same predicate `FlushResolvedCache` already uses (dns_redirect.go:122) — so we never invent names on a host whose resolver wouldn't have answered them either. A/AAAA and the matching NODATA for other qtypes; short TTL, since the route can change mid-run.

**2. Don't let the filter gate refuse them.**

Answering before the gate covers this, and is better than an allow-rule carve-out: these names are locally synthesized, never leave the box, and carry no tunneling channel.

**3. Log the NXDOMAIN case until (1) lands.**

An upstream NXDOMAIN for a name we know is synthetic is a CargoWall-caused failure and should not be silent in audit mode.

## Out of scope

- **Auto-allowing the synthesized address.** Resolution and egress are separate decisions. On Blacksmith the gateway fronts a metadata service that vends `stickyDiskToken`; granting reachability because we answered a DNS query would be a policy change smuggled in through a resolver fix. Operators allow it explicitly — `192.168.127.0/24`, all ports, in Blacksmith's case.
- **Making `_gateway` writable as a hostname rule.** The control plane's `BeValidHostname` regex has no underscore in its DNS label class, so `_gateway` is refused on every write path. That is correct — a leading underscore is not a legal hostname label, which is exactly why systemd chose the name — and CIDR is the right allow path regardless. Separate concern, no change requested here.
- **Undoing the stub DNAT.** The attribution reason for it stands.

matthewdevenny:
Closes #126

## Problem

systemd-resolved answers `_gateway`, `_outbound`, `_localdnsstub` and `_localdnsproxy` from local state and never sends them upstream. The DNS redirect DNATs the resolved stub to the proxy, and the proxy's upstream is deliberately the resolver *behind* resolved, so every one of those names NXDOMAINed for the length of a run — REFUSED under enforce, silently under audit, where no connection is ever attempted and nothing is logged. `localhost` survives only because nsswitch consults `/etc/hosts` first.

Found via Blacksmith runners, whose host-side agent (sticky disks, cache, metadata) is reached at the VM's default gateway; a runner generation that names it `_gateway` loses the agent entirely, and the Blacksmith actions fall back silently while CargoWall reports nothing.

## Change

**`pkg/dns/synthetic.go`** — the proxy relays the names to the resolved stub on its already-marked client socket (`s.client`; the redirect's RETURN rules pass the mark through the stub DNAT, exactly as they do for the startup cache peek) and writes resolved's own answer back: exact rcode, flags and TTL, uncached, unenforced. Covers the underscore set and the RFC 6761 localhost family (`localhost`, `localhost.localdomain`, anything beneath either — `/etc/hosts` carries only the exact name `localhost`). No route-table parsing, no source-address heuristic, no new dependencies.

Gates, in order: IN class; host listener (a container's native path never resolved these, and the host's gateway is the wrong answer in a container netns); name matches. Sits ahead of the filter gate and the cache — resolved's own precedence. There is deliberately no "is resolved running" probe: `/run/systemd/resolve` outlives a stopped resolved (`RuntimeDirectoryPreserve=yes`), so directory presence proves nothing; a stub that does not answer sends the query down the ordinary path at the cost of one connection-refused round trip on loopback, and a resolved stop or restart mid-run is honoured with no state.

**Search-expanded form.** With a search list in `resolv.conf`, every stub resolver asks `_gateway.<search>` *before* `_gateway`, and the gate REFUSED it. glibc and Go treat REFUSED as "try the next form"; c-ares ends its search on a REFUSED multi-label attempt (its #852 carve-out is single-label only), so `_gateway` was unresolvable for c-ares clients under enforce even with the bare name answered. That form is answered NXDOMAIN locally — what the upstream says natively — with no round trip and no block record. Only suffixes on the host's own `resolv.conf` search list qualify; `_gateway.example.com` stays an ordinary query. It is never relayed to the stub, which would send it upstream unmarked and DNAT it straight back.

**Resolution only, never egress.** The answer feeds neither hostname tracking nor the firewall; the test server carries a mock firewall with zero expectations, so any enforcement side effect fails the suite. Reaching the gateway stays an explicit CIDR rule, per the issue's out-of-scope.

`pkg/dns/server.go` gains one `serveSynthetic` early return; everything else lives in the new file. `pkg/network` is untouched.

## Verification

Lima VM (CI's 6.17-azure kernel), live under enforce with query filtering and `search lan`, through the real DNAT:

| Client | Before | After |
|---|---|---|
| c-ares (pycares) `_gateway` | `error=6 DNS server refused query` | `192.168.5.2` |
| glibc `getent hosts _gateway` / `_outbound` | NXDOMAIN | `192.168.5.2` / `192.168.5.15` |
| Go pure / cgo resolver | NXDOMAIN | resolves |
| `dig @127.0.0.53 _gateway` (stub, DNAT path) | NXDOMAIN | resolved's own answer: `aa ra ad`, TTL 0 |
| `_gateway.lan` on the wire | REFUSED + block record per lookup | NXDOMAIN `aa`, no record |
| `api.localhost` via the stub | REFUSED | `127.0.0.1` |
| resolved stopped mid-run (`systemctl stop`) | — | `_gateway` fails quietly on the ordinary path, no WARN; works again on start with no cargowall restart |
| `example.com` control | REFUSED | REFUSED |

Every run tears the iptables rules down cleanly. 10 new unit tests (stub faked with a local `dns.Server`, started before use and closed on cleanup); full `go test ./...` green in Lima; `build`, `vet`, `staticcheck`, `gofumpt`, `goimports-reviser` clean.

## Follow-up

#127 — the same REFUSED-vs-c-ares failure hits any *allowed* single-label hostname on a host with a search domain (rule `myhost`, client asks `myhost.lan` first). Same fix shape; kept separate because it touches gate semantics rather than resolved fidelity.


pkg/dns/server.go
```diff
@@ -385,6 +390,10 @@ func (s *Server) Start(ctx context.Context) error {
 	// No TTL cleanup needed - IPs persist until updated by new DNS responses
 	// DNS cache uses lazy expiration - no cleanup goroutine needed
 
+	// Decide ahead of the listeners whether to answer resolved's synthetic
+	// names (#126); written here and never again.
+	s.synthetic = s.probeSynthetic()
```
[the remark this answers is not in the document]
matthewdevenny: Fixed in de27956. The Start-time snapshot is gone: `serveSynthetic` calls `network.SystemdResolvedRunning` per hit — only for queries whose first label is a synthetic name, so it is one `stat` on a rare path — and a resolved stop or restart mid-run is honoured immediately. A probe error reads as not running and the query takes the ordinary path.

pkg/dns/synthetic.go
```diff
@@ -0,0 +1,402 @@
...
+			ipn, ok := a.(*net.IPNet)
+			if !ok || (ipn.IP.To4() != nil) != v4 {
+				continue
+			}
+			if ipn.Contains(rt.gateway) {
+				pick = ipn.IP
+				break
+			}
+			if fallback == nil && ipn.IP.IsGlobalUnicast() {
+				fallback = ipn.IP
```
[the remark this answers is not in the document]
matthewdevenny: Resolved by removing the code in de27956. The proxy no longer derives `_outbound` (or `_gateway`) itself: bare synthetic names are relayed to the resolved stub on the proxy's already-marked client socket, and resolved's own answer — its kernel route lookup, preferred source and all — is written back verbatim, uncached and unenforced. The `/proc` parsers and the subnet heuristic are gone.

matthewdevenny:
Two of Copilot's suppressed findings, for the record (no thread to reply in):

- **Qclass** — taken in de27956: `serveSynthetic` intercepts IN class only; a CHAOS `_gateway` query takes the ordinary path (pinned by `TestServeSynthetic_NonINClassIsOrdinary`).
- **Suffix-only detection could shadow a real `_gateway.<search>` upstream** — declined. Under enforce that record is unreachable either way: no rule can carry an underscore label, so the gate REFUSES it, and this change only swaps that REFUSED for the NXDOMAIN c-ares needs in order to move on to the bare name. Under audit it is a deviation only for a zone that actually defines `_gateway.<your search domain>`. Relaying the expanded form to the stub is not an option — resolved would send it upstream unmarked and the DNAT would bring it straight back into the proxy. Documented in design.md.

pkg/dns/synthetic.go
```diff
@@ -0,0 +1,133 @@
...
+// resolved running (probed per hit) — reporting whether it wrote a response.
+func (s *Server) serveSynthetic(w dns.ResponseWriter, r *dns.Msg) bool {
+	if len(r.Question) == 0 || r.Question[0].Qclass != dns.ClassINET || !s.hostListener(w) {
+		return false
+	}
+	name, expanded, ok := syntheticQuery(r.Question[0].Name)
+	if !ok {
+		return false
+	}
+	if running, err := resolvedRunning(); err != nil || !running {
```
jordanjennings: `SystemdResolvedRunning` doesn't detect a stopped resolved: `systemd-resolved.service` sets `RuntimeDirectoryPreserve=yes` ([units/systemd-resolved.service.in](https://github.com/systemd/systemd/blob/main/units/systemd-resolved.service.in), same in v249), so `/run/systemd/resolve` survives `systemctl stop systemd-resolved`.

A job that stops resolved (e.g. to run dnsmasq or kind DNS) then has every `_gateway`/`_outbound` lookup relayed to a dead stub → SERVFAIL (EAI_AGAIN, which retry loops keep hammering) plus a `Warn` per attempt, which the Actions handler emits as a `::warning::` annotation. Before this PR those lookups failed quietly. The per-hit probe covers a restart but not a stop.

Suggest falling through to the ordinary path (`return false`, Debug log) when the stub exchange errors.

pkg/dns/synthetic.go
```diff
@@ -0,0 +1,133 @@
...
+		if err != nil {
+			s.logger.Debug("systemd-resolved probe failed; synthetic name takes the ordinary path",
+				"name", name, "error", err)
+		}
+		return false
+	}
+
+	m := new(dns.Msg)
+	if expanded {
```
jordanjennings, line 59: Design question: the c-ares behaviour this works around (REFUSED on a search-expanded attempt terminates the search) isn't specific to these four names. It hits any allowed single-label name whose host search suffix isn't in cargowall's `SearchDomains`, and those come from cloud-provider detection (`cmd/start.go:1402`), not the host's `resolv.conf`, so self-hosted runners and GCP project-scoped search lists are exposed.

E.g. rule allows `myservice`, host has `search corp.lan`: c-ares tries `myservice.corp.lan`, gets REFUSED, never tries `myservice`.

Would a general fix in `handleDNSQuery`'s blocked branch (NXDOMAIN instead of REFUSED for `<label>.<host search suffix>`) cover this and the synthetic case together?

pkg/dns/synthetic.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/miekg/dns"
+
+	cargowallNet "github.com/code-cargo/cargowall/pkg/network"
+)
+
+// syntheticNames are the names systemd-resolved synthesizes locally (#126).
+// The proxy relays them to the stub on its marked client and writes
+// resolved's answer back uncached and unenforced: it never feeds
+// hostnameIPs or the firewall.
+var syntheticNames = []string{"_gateway", "_outbound", "_localdnsstub", "_localdnsproxy"}
```
jordanjennings: resolved also synthesizes `*.localhost` and `localhost.localdomain` (→ 127.0.0.1 / ::1). `/etc/hosts` only covers the exact name `localhost`, so a client resolving through the stub (e.g. glibc with `files dns`) that looks up `api.localhost` under enforce still gets REFUSED, the same failure as #126. Either cover these here or narrow the design.md scope note.

pkg/dns/synthetic_test.go
```diff
@@ -0,0 +1,299 @@
...
+				m.Answer = append(m.Answer, &dns.A{
+					Hdr: dns.RR_Header{Name: q.Name, Rrtype: dns.TypeA, Class: dns.ClassINET, Ttl: 0},
+					A:   net.ParseIP(ip).To4(),
+				})
+			}
+			_ = w.WriteMsg(m)
+		}),
+	}
+	go func() { _ = stub.ActivateAndServe() }()
+	t.Cleanup(func() { _ = stub.Shutdown() })
```
jordanjennings: `Shutdown`'s error is discarded, and in miekg/dns v1.1.73 `ShutdownContext` returns "server not started" without closing the conn if `ActivateAndServe` hasn't run yet (`server.go:413`), the same pitfall `failStart` in `server.go` already works around. Tests that never touch the stub (`ExpandedIsNXDomain`, `ResolvedNotRunningIsOrdinary`, `NonINClassIsOrdinary`, `ContainerListenerIsOrdinary`) can finish first, leaking a UDP socket and a goroutine that then serves forever. Close `pc` in the cleanup, or wait on `NotifyStartedFunc` before returning.

pkg/dns/synthetic.go
matthewdevenny: Right, and verified on the Lima box: `RuntimeDirectoryPreserve=yes` is in the unit, `/run/systemd/resolve` is still there after `systemctl stop systemd-resolved`, and the stub returns connection refused. So the probe only ever detected "started once since boot".

Fixed in 2b6ff22 by taking your suggestion and going one step further: the probe is gone entirely (and `pkg/network` is back to main — the extraction was only there to share it). A stub exchange error now `return false`s with a Debug line, so the name takes the ordinary path, which is where it went before this PR. Cost is one connection-refused round trip on loopback per synthetic hit on a host without a live stub, which is the cheapest true probe available.

Live under enforce: stop resolved mid-run → `_gateway` is REFUSED at the gate, `getent` rc=2, c-ares refused, zero WARN lines; `systemctl start` → answers again with no cargowall restart.

pkg/dns/synthetic.go
matthewdevenny: Agreed the general problem is real, and your `myservice` / `search corp.lan` example is exactly the shape — and it also shows why the general fix is not this one.

For a real host, the expanded form is the name that *exists*: `myservice.corp.lan` is the record, `myservice` is the convenience the search list provides. NXDOMAINing the expanded form and letting the client fall through to bare `myservice` would break the ordinary enterprise short-name pattern rather than fix it. What the general case needs is the opposite: strip the host's `resolv.conf` search suffixes before the gate judges the name — the Kubernetes-suffix semantics `StripSearchDomains` already has — so `myservice.corp.lan` is matched as `myservice` and forwarded when allowed, and the client gets the real answer on its first attempt. Stripping only, not the search-domain bypass, since `resolv.conf` is DHCP-written.

NXDOMAIN is right only for the synthetic set, whose expanded form never exists and whose bare form is what resolves. So the special case stays here, and #127 — which I had drafted with NXDOMAIN, wrongly — is corrected to the stripping proposal. It also now names the gap you point at: cargowall's `SearchDomains` come from cloud-provider detection, not the host's `resolv.conf`.

pkg/dns/synthetic.go
matthewdevenny: Covered in 2b6ff22. Verified on the box first: `api.localhost`, `localhost.localdomain` and `foo.localhost.localdomain` all come back `Data from: synthetic`, and `/etc/hosts` carries only the exact name `localhost`. The relay now matches `localhost`, `localhost.localdomain` and anything beneath either, and hands them to the stub like the underscore set. Live under enforce through the stub path all three answer `127.0.0.1`. design.md scope note updated to list the family.

pkg/dns/synthetic_test.go
matthewdevenny: Fixed in 2b6ff22. `withFakeStub` now blocks on `NotifyStartedFunc` before returning, so `Shutdown` in cleanup always runs against a started server, and the cleanup closes `pc` as well. Checked the v1.1.73 source you cite — `ShutdownContext` returns `"server not started"` before touching the conn — so the leak was real for the four tests that never query the stub.

jordanjennings (approved):

PR #131, opened 2026-09-15
#129 enforce the relayed synthetic answer so `_gateway` can be a rule

### LINKED ISSUE #129: Allow underscore labels in hostname rules so `_gateway` can be allowed by name, not by provider CIDR
## Symptom

Underscore labels are legal DNS (RFC 2181 §11; `_dmarc`, `_acme-challenge`, `_sip._tcp`, RFC 8552) and modern resolvers accept A records at them, but a hostname rule cannot carry one: the control plane's `DnsLabel` regex (`app`: `CargoWallAiRuleValidators.cs`, `[a-zA-Z0-9]([a-zA-Z0-9\-]{0,61}[a-zA-Z0-9])?`) is RFC 1123 hostname syntax, and every write path — policy editor, tag rules, AI rules, `cargowall_add_allow_rule` — goes through it. Two things fall out of that:

1. A legacy internal host with an underscore in its name (`build_server.corp.example`) cannot be allowed at all.
2. systemd-resolved's synthetic names cannot be allowed by name. #126 / #128 make `_gateway` and friends *resolve* again by relaying them to the resolved stub, but deliberately as resolution only: the relayed answer feeds neither hostname tracking nor the firewall. So for a runner whose host agent sits at the default gateway (Blacksmith: `BLACKSMITH_AGENT_ADDR=192.168.127.1`, sticky disks / cache / metadata on per-VM ports), the operator sees a block on a bare `192.168.127.1:1041` with no name attached, and the only fix is a CIDR that is a gvproxy implementation detail of one runner generation.

The agent side already accepts underscore rule values — `validateRules` checks only ICMP/port constraints — so the gap is the control-plane regex plus the relay's "never egress" boundary.

## Why the boundary is where it is

#128 kept the relay out of enforcement on purpose: answering a name is not the same decision as granting egress to what it resolves to, and a route-derived allow with no rule behind it would hand every job the host. That reasoning holds. What it does not require is that the name be *un-allowlistable* — only that allowing it be an explicit rule, like any other hostname.

## Proposal

Ship as one feature; the halves are useless apart. Attribution without allowlistability is a dead end (the log says `_gateway`, the operator adds a `_gateway` rule, the control plane refuses it), and allowlistability without enforcement is a rule that validates and does nothing.

**1. Control plane (app): accept underscores in hostname rule labels.** `DnsLabel` → `[a-zA-Z0-9_]([a-zA-Z0-9_\-]{0,61}[a-zA-Z0-9_])?` or equivalent. Leading underscore included — that is the whole point of the RFC 8552 convention. Keep everything else: label ≤ 63, name ≤ 253, not an IP, wildcard segments whole. `CargoWallRuleLint` is unaffected. Update the `cargowall_add_allow_rule` description, which currently says "hostnames".

**2. Agent (`pkg/dns/synthetic.go`): route the relayed answer through enforcement.** After the stub `Exchange`, pass the response through `enforceDNSResponse` exactly as an upstream answer is, skipping the cache (TTL 0 anyway). Then:

- with a `_gateway` allow rule, the gateway address(es) are opened on the rule's ports — `allowed-hosts: _gateway` replaces the per-provider CIDR and survives a runner-generation change;
- without one, the block is attributed — `dst=_gateway dst_ip=192.168.127.1` — instead of an unattributed IP;
- a `_gateway` **deny** rule becomes expressible, which is a real hardening primitive.

Resolution stays ahead of the filter gate (a local answer carries no tunneling channel), so under enforce the name resolves and the *connection* is judged — better diagnostics than a REFUSED query. Suffix matching is unchanged: `_gateway` matches `_gateway` and `x._gateway`, never `_gateway.lan`, so the search-expanded NXDOMAIN path in #128 is untouched. `_outbound`, `_localdnsstub` and `_localdnsproxy` come along for free (loopback is auto-allowed already).

**3. Docs.** One sentence on `allowed-hosts`: a `_gateway` rule trusts the routing table the way a hostname rule trusts DNS. Recommend repository/workflow scope, not org — at org level it means "every runner's default gateway, whatever that is".

## Out of scope

- Auto-allowing the gateway. On Blacksmith it fronts a metadata service that vends `stickyDiskToken`; reachability stays an explicit rule.
- Relaxing `validateSearchDomain` (agent) or `SearchDomainRegex` (app) for underscores. Search domains are suffixes appended to lookups; nothing needs an underscore there.
- #127 (search-expanded forms of allowed single-label names). Independent; `_gateway` under this issue would still hit it for c-ares only via the #128 NXDOMAIN path, which already covers the synthetic set.

matthewdevenny:
Part of #129 — the agent half. The control-plane half (admitting underscore labels in `CargoWallValidationHelpers.BeValidHostname`, plus the `cargowall_add_allow_rule` description and a unit test class) is a companion change in `app`, left unstaged there for review per that repo's conventions. #129 closes when both land.

## Problem

#128 made `_gateway` resolve again, but deliberately as resolution only: the relayed answer fed neither hostname tracking nor the firewall. So on a runner whose host agent sits at the default gateway (Blacksmith: `BLACKSMITH_AGENT_ADDR=192.168.127.1`) the operator sees a block on a bare IP with no name attached, and the only fix is a CIDR that is one runner generation's gvproxy detail.

## Change

**Which synthetic names are enforced** — one classifier, `lookupSynthetic` → (name, `aliasClass`: `notAlias` / `trackedAlias` / `untrackedAlias` / `expandedAlias`), checked in place against the fixed sets with nothing built per query:

- `_gateway`, `_outbound` and the **machine's own hostname** (its configured name and, when it carries a domain, its first label — read per lookup, so a `hostnamectl set-hostname` mid-run is honoured; `/etc/hosts` does not always pin it, and under the redirect `getent hosts $(hostname)` failed and `sudo` warned on every invocation) get **L4 enforcement and attribution** like an upstream answer: an allow rule opens the address on its ports, a deny closes it, no rule → the block is attributed to the name instead of a bare IP.
- `_localdnsstub`, `_localdnsproxy` and the localhost family are **relay-only**: loopback listeners the relay itself depends on, and any label under `.localhost` is accepted, so tracking them would grow `hostnameIPs` without bound.
- The search-expanded form a stub resolver tries first (`<name>.<search>`) is NXDOMAIN for every one of these, as #128 did for `_gateway` — REFUSED ends a c-ares search before the bare name. The mDNS `<label>.local` form is not relayed: resolved renames it on conflict, and an unsynthesized `.local` query is multicast to the LAN.

**No L7 identity for an alias.** `_gateway` is an address alias, never a name a peer presents; binding it would L7-scope the address against a name no flow can show, and an all-ports allow would then drop HTTP to the host agent for carrying `Host: 192.168.127.1`. Two paths mint forward-resolution evidence and both now withhold it for such names:

- `enforceDNSResponse` takes an `answerSource` — `wireAnswer` from the upstream and CNAME pre-resolve callers, `localAnswer` from the relay. The shared function knows nothing of synthetic names.
- Startup pre-population resolves every tracked rule name through the stub and now records through `dns.Server.RecordSystemCacheAnswer`, so `cmd` restates no policy: a wire name is mapped and mints evidence, a tracked alias is mapped without evidence, a loopback listener is recorded nowhere — so the tracked-hostname replay can never write the stub's own address. Without this, a `_gateway` rule plus `--prepopulate-dns-cache` (both presets) L7-scoped the gateway regardless of what the relay did.

SCOPE IFF BOUND makes the omission sufficient; nothing in the L7 matcher learns about synthetic names. The rule is L4-only, like the CIDR it replaces.

## Verification

Lima, enforce + query filtering, `search lan`, `curl 192.168.5.2:8080` (the VM's gateway, nothing listening):

| | no rule | `_gateway` allow rule |
|---|---|---|
| `getent hosts _gateway` | `192.168.5.2` | `192.168.5.2` |
| proxy log | `Connection blocked … dst=_gateway dst_ip=192.168.5.2 dst_port=8080` | `DNS resolution: added to firewall hostname=_gateway … action=allow`, then `Connection allowed … dst=_gateway` |
| curl | timeout — SYN dropped | rc=7, refused by the peer — the SYN went out |
| machine hostname: c-ares / glibc / `sudo` | resolves / resolves / no warning | — |
| `<hostname>.lan` on the wire | NXDOMAIN | — |

Unit (`pkg/dns`): relay tracked-not-cached with no firewall call absent a rule; allow → `AddIP(192.168.5.2, allow, tcp/1041)`, deny under default allow → `AddIP(…, deny)`; `AllowDoesNotL7Scope` (no evidence, registrar untouched, attribution intact) with the wire-path contrast (`myservice.corp.lan` mints evidence, `all-ports` scope); `StubListenersUntracked`; `MachineHostnameRelayedAndTracked` incl. expanded-form NXDOMAIN and a mid-run rename; `TestMachineNames` incl. localhost screening; `TestRecordSystemCacheAnswer` on the pre-population path (all three classes); `TestLookupSynthetic` (25 cases). The Lima checks are L4-only (`--tls-sni` off; nothing listens at the gateway on an HTTP port): the L7 seam is pinned by the unit tests plus `TestRegisterL7ScopesOnlyBoundIPs`. Full suite green; vet, staticcheck, gofumpt, goimports-reviser clean.

## Docs

README (`allowed-hosts`): underscore labels are valid; `_gateway` names the runner's default gateway; the rule is L4-only and trusts the routing table, scope it to repo/workflow; it needs systemd-resolved — without a stub the name does not resolve and the rule opens nothing. design.md's synthetic-names bullet and design-l7.md's pre-population paragraph agree on evidence.

## Not done here

Auto-allowing the gateway — reachability stays an explicit rule. Recommending `_gateway` at org level. Residual recorded in design.md: a machine hostname that is also a real DNS name resolves to resolved's synthesized address rather than the upstream record — inherent to making the machine hostname resolve, and what the host does natively.


pkg/dns/synthetic.go
```diff
@@ -68,6 +70,9 @@ func (s *Server) serveSynthetic(w dns.ResponseWriter, r *dns.Msg) bool {
 		return false
 	}
 	resp.Id = r.Id
+	if resp.Rcode == dns.RcodeSuccess {
+		s.enforceDNSResponse(name, resp, 0)
+	}
```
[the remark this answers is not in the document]
matthewdevenny: Fixed in 1604e8b. Enforcement now applies to the four underscore names only — a fixed set, so tracking stays bounded — and the localhost family goes back to relay-only as in #128: loopback is auto-allowed, and attributing 127.0.0.1 to an alias buys nothing. `TestServeSynthetic_RelaysLocalhostFamilyUntracked` pins that `api.localhost` answers but leaves `hostnameIPs` and the IP→name map empty; the `_gateway` allow/deny/attribution tests are unchanged. For the record, the same no-eviction property already holds for any name an allow rule admits (a `**.example.com` rule tracks every label a job invents), so the delta here was specifically that the localhost family needed no rule to be tracked.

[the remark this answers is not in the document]
matthewdevenny: Final shape as of 2c7cf6d: enforcement (and therefore tracking) applies only to the underscore set and the machine's own hostname forms — a fixed handful of names — and the localhost family is relay-only and never tracked, so `hostnameIPs` cannot grow from arbitrary `*.localhost` labels. Classification lives in one place (`lookupSynthetic`), and `TestServeSynthetic_RelaysLocalhostFamilyUntracked` pins that `api.localhost` answers while the IP→name map and `hostnameIPs` stay empty. Resolving.

pkg/dns/synthetic.go
```diff
@@ -68,6 +132,9 @@ func (s *Server) serveSynthetic(w dns.ResponseWriter, r *dns.Msg) bool {
 		return false
 	}
 	resp.Id = r.Id
+	if enforce && resp.Rcode == dns.RcodeSuccess {
+		s.enforceDNSResponse(name, resp, 0, false)
```
jordanjennings: `wireIdentity=false` keeps this path from minting L7 evidence, but the startup pre-population path mints it anyway.

`prePopulateDNSCache` iterates `GetTrackedHostnames()` — which includes `_gateway` once a hostname rule names it (`pkg/config/config.go:2196`) — and calls `configMgr.RecordForwardResolution(hostname, ip)` directly (`cmd/start.go:1619-1629`). The lookup goes to the stub at `127.0.0.53` and Go's `isDomainName` accepts underscores, so it succeeds. `ApplyRulesToTrackedHostnames` then replays the IP through `registerL7`, the SCOPE-IFF-BOUND check at `pkg/dns/l7.go:101` passes, and the gateway address gets L7-scoped. A subsequent `curl http://192.168.127.1:...` presents `Host: 192.168.127.1`, matches no hostname rule, and is denied — precisely what README.md:107 promises cannot happen.

Preconditions: `--prepopulate-dns-cache` (both CI presets set it) plus `--tls-sni=enforce` and `--container-egress=enforce`. At the default `--tls-sni=off`, `l7Registrar()` is nil and `ScopeIP` is a no-op, so this stays invisible.

Nothing excludes synthetic names from `GetTrackedHostnames()`, and no test covers the cmd path.

pkg/dns/synthetic.go
```diff
@@ -17,24 +17,88 @@
...
+}
+
+// lookupSynthetic classifies a wire-form query name (trailing dot). ok:
+// the proxy answers this name itself. expanded: the search-expanded form of
+// an underscore name, answered NXDOMAIN — never relayed, never enforced.
+// enforce: the answer gets L4 enforcement and attribution like an upstream
+// one (#129) — the underscore set and the machine's own names, a fixed
+// handful. The localhost family is relayed only: any label under .localhost
+// is accepted, and tracking each alias would grow hostnameIPs without bound.
+// Machine names take no search expansion: a real host's expanded form is
```
jordanjennings: This premise does not hold under query filtering, and the case where it fails is the one the seeding exists for.

On a host with `search vm.blacksmith.sh`, `sudo` calls `getaddrinfo("runner-abc")`; 0 dots < ndots:1, so glibc tries `runner-abc.vm.blacksmith.sh` first. Machine names take no expansion, so that falls to the ordinary path and the filter gate answers REFUSED (`server.go:569`). glibc maps REFUSED to `NO_RECOVERY`, which hits the `default: done++` arm of `res_nsearch` — the bare `runner-abc` that `serveSynthetic` would have answered is never sent. c-ares aborts identically, which is the same reason this PR answers `_gateway.<search>` NXDOMAIN rather than REFUSED.

`synthetic_test.go:392-393` pins that REFUSED as correct behavior, and the fixtures at `synthetic_test.go:150` and `:310` use `vm.blacksmith.sh` as the exemplar search domain. The premise holds only when `<hostname>.<search>` exists upstream AND a rule allows it; under query filtering neither is true for a runner whose hostname bears no relation to the DHCP search domain.

pkg/dns/synthetic.go
```diff
@@ -17,24 +17,88 @@
...
+	first, _, cut := strings.Cut(h, ".")
+	if first == "" {
+		return
+	}
+	add := func(n string) {
+		if !slices.Contains(s.machineNames, n) {
+			s.machineNames = append(s.machineNames, n)
+		}
+	}
+	add(h)
```
jordanjennings: `seedMachineNames` does not screen the hostname against `localhostRoots`. On an unconfigured host where `os.Hostname()` returns `localhost`: `first="localhost"`, `cut=false`, so `machineNames` becomes `["localhost", "localhost.local"]`.

`lookupSynthetic("localhost.")` is caught by the `localhostRoots` loop (`enforce=false`, correct), but `"localhost.local"` matches neither root — it ends in `.local`, not `.localhost` — so it falls through to the `machineNames` check at line 87 and returns `enforce=true`. `enforceDNSResponse` then writes `ipToHostname["127.0.0.1"]` and `hostnameIPs`, which is exactly what `TestServeSynthetic_RelaysLocalhostFamilyUntracked` (`synthetic_test.go:299-302`) asserts must not happen.

pkg/dns/synthetic.go
```diff
@@ -17,24 +17,88 @@
...
+		}
+	}
+	if slices.Contains(s.machineNames, full) {
+		return full, false, true, true
+	}
+	first, rest, hasRest := strings.Cut(full, ".")
+	if !slices.Contains(syntheticNames, first) {
+		return "", false, false, false
+	}
+	if !hasRest {
```
jordanjennings: This returns `enforce=true` for `_localdnsstub` and `_localdnsproxy`, so a rule naming either one resolves it to `127.0.0.53`/`127.0.0.54` and writes BPF entries plus `ipToHostname["127.0.0.53"] = "_localdnsstub"` — the loopback tracking the localhost-family branch above was written to avoid. These are the stub addresses the proxy's own relay and startup cache peek depend on.

`CheckIPRuleConflict` skips every non-CIDR rule (`pkg/config/config.go:1506-1509`), so a loopback auto-allow that is not a CIDR rule will not veto a deny here. Only `_gateway` is exercised by the tests.

pkg/dns/synthetic.go
```diff
@@ -17,24 +17,88 @@
...
+// Machine names take no search expansion: a real host's expanded form is
+// the name that resolves, so "runner.lan" stays on the ordinary path.
+func (s *Server) lookupSynthetic(qname string) (name string, expanded, enforce, ok bool) {
+	full := strings.ToLower(strings.TrimSuffix(qname, "."))
+	for _, root := range localhostRoots {
+		if full == root || strings.HasSuffix(full, "."+root) {
+			return full, false, false, true
+		}
+	}
+	if slices.Contains(s.machineNames, full) {
```
jordanjennings: This intercepts the machine's configured FQDN unconditionally, and `serveSynthetic` runs at `server.go:541`, ahead of the filter gate at `:548`.

On a host whose `hostname` is a real DNS name (`runner.corp.example.com`) that internal DNS serves with a different address — a VIP or a service-registry record — the query no longer reaches `s.upstream`; it returns resolved's synthesized local address instead. A workload resolving its own FQDN for service discovery gets the wrong answer, a policy that explicitly denies that hostname no longer prevents it from resolving, and the answer is never cached, so every lookup costs a stub round trip.

pkg/dns/synthetic.go
```diff
@@ -17,24 +17,88 @@
...
+	resolvedStubAddr = "127.0.0.53:53"
+	machineHostname  = os.Hostname
+)
+
+// seedMachineNames records the forms resolved synthesizes for the machine's
+// own hostname — the configured name, its first label when it carries a
+// domain, and the mDNS form "<label>.local" — so they resolve under the
+// redirect even where /etc/hosts does not pin them (sudo's "unable to
+// resolve host" otherwise). Deduplicated; three names at most.
+func (s *Server) seedMachineNames() {
```
jordanjennings: `seedMachineNames` is called only from `Start`, so the hostname is captured once and never refreshed. After a `hostnamectl set-hostname` mid-run — a provisioning script racing cargowall's start, or a workflow step — the stale names stay intercepted (resolved no longer synthesizes them, so it answers not-found), while the current one falls to the ordinary path and is REFUSED under query filtering. `sudo` warns exactly as it did before the fix, and there is no re-seed hook.

pkg/dns/synthetic.go
```diff
@@ -17,24 +17,88 @@
...
+	add := func(n string) {
+		if !slices.Contains(s.machineNames, n) {
+			s.machineNames = append(s.machineNames, n)
+		}
+	}
+	add(h)
+	if cut {
+		add(first)
+	}
+	add(first + ".local")
```
jordanjennings: This guesses resolved's mDNS hostname by string concatenation. resolved renames on mDNS conflict (`runner-abc-2.local`), and `m->mdns_hostname` is also stale if the hostname changed after resolved started. In either case `runner-abc.local` is not synthesized — resolved routes it to the `.local` mDNS scope and multicasts it on the LAN.

`lookupSynthetic` still returns `enforce=true` for it, so whatever a LAN peer answers is fed to `enforceDNSResponse` and `UpdateDNSMapping`, and if any rule names that form, `applyVerdictSide` opens the responder's address. The premise quoted in the design bullet, "an answer the proxy produced from resolved's local state", is false for the `.local` form.

README.md
```diff
@@ -104,6 +104,8 @@ allowed-hosts: |
   **.storage.azure.com
 ```
 
+Underscore labels are valid rule values. `_gateway` names whatever systemd-resolved resolves it to — the runner's default gateway — so a host-side agent reached at the gateway (Firecracker/gvproxy runners such as Blacksmith put theirs at `192.168.127.1`) can be allowed by name rather than by a provider-specific CIDR. Two things to know: such a rule trusts the routing table the way any hostname rule trusts DNS, so scope it to the repository or workflow that needs it, not the organization; and `_gateway` is an address alias, never a name the peer presents, so the rule is L4-only — it opens the address on its ports like a CIDR allow and pins no L7 identity, so with `--tls-sni` on, HTTP to the agent carrying `Host: 192.168.127.1` is not an L7 miss.
```
jordanjennings: This tells operators to replace a provider-specific CIDR with `_gateway`, but does not mention that the feature requires systemd-resolved.

Without a resolved stub — not installed, or stopped — `s.client.Exchange(r, resolvedStubAddr)` fails, the query takes the ordinary path, `_gateway` NXDOMAINs upstream (or is REFUSED under filtering), no IP is ever tracked, and traffic to the host agent is blocked. That is a silent regression relative to the CIDR being replaced. The trade-off is framed here only as "trusts the routing table"; design.md documents the fall-through, the README does not.

pkg/dns/server.go
```diff
@@ -660,8 +665,12 @@ func (s *Server) handleDNSQuery(w dns.ResponseWriter, r *dns.Msg) {
...
 // canonicalHostname is the queried name, lowercased with the trailing dot
 // trimmed. depth bounds pre-resolve recursion (see preResolveCNAMETarget);
-// handleDNSQuery passes 0.
-func (s *Server) enforceDNSResponse(canonicalHostname string, resp *dns.Msg, depth int) {
+// handleDNSQuery passes 0. wireIdentity reports whether the answer came off
+// the wire — an upstream or pre-resolve answer for a name a peer can
+// present — and so may mint L7 forward-resolution evidence; an answer the
+// proxy produced from resolved's local state (serveSynthetic) is an address
+// alias no peer presents and mints none, so SCOPE IFF BOUND scopes nothing.
+func (s *Server) enforceDNSResponse(canonicalHostname string, resp *dns.Msg, depth int, wireIdentity bool) {
```
jordanjennings: Nit: `wireIdentity bool` guards a security invariant with a positional bool, and two of the three call sites pass a bare `true`. Nothing stops a future caller from defaulting to `true` and silently minting binding evidence for a locally-produced answer.

A named type (`sourceWire`/`sourceLocal`), or hoisting `RecordForwardResolution` out into the two wire callers, would make the omission structural rather than a per-call-site convention — the same argument `RegisterL7Identity`'s doc comment makes for enforcing SCOPE IFF BOUND centrally "rather than per caller".

jordanjennings (changes requested):
Blocking on two issues: the startup pre-population path defeats the `wireIdentity=false` invariant this PR introduces (`pkg/dns/synthetic.go:136`), and the machine-name seeding does not fix `sudo` on a host with a resolv.conf search list, which is the environment it exists for (`pkg/dns/synthetic.go:78`).

One cross-cutting note: `design.md:701` and `design-l7.md:286-289` now contradict each other on whether startup pre-population mints forward-resolution evidence. Production follows design-l7.md.

pkg/dns/synthetic.go
matthewdevenny: Fixed in c22ba83. Phase-1 recording is now `recordSystemCacheAnswer` in `cmd/prepopulate.go`: it asks `dns.Server.LocalAlias` (the proxy's own classification, `lookupSynthetic`) and maps such names for attribution without minting evidence, so `ApplyRulesToTrackedHostnames` has nothing to scope. `TestRecordSystemCacheAnswer_WithholdsEvidenceForLocalAliases` pins both directions on the cmd path (`_gateway` → mapped, no evidence; `registry.example` → mapped, evidence). design-l7.md now carries the exception in its pre-population paragraph and design.md says the same thing, so the contradiction is gone. I did not add a live L7 check — nothing listens at the gateway on an HTTP port in Lima — so the seam is pinned by that test plus `TestRegisterL7ScopesOnlyBoundIPs`.

pkg/dns/synthetic.go
matthewdevenny: Fixed in c22ba83: the machine's names take the same search-expansion path as `_gateway` — `<hostname>.<search>` is answered NXDOMAIN, never relayed — pinned in `TestLookupSynthetic` and `TestServeSynthetic_MachineHostnameRelayedAndTracked`. Live under enforce with `search lan`, c-ares (pycares) now resolves the machine hostname where it was refused before.

One correction on the analysis, since it decides who this bites: glibc does fall through on REFUSED. On this box, with `search lan` and the previous commit's REFUSED for `lima-ebpf-devel.lan`, `getent hosts lima-ebpf-devel` returned the address and `sudo -n true` printed no warning — `res_nsearch` tries the name as-is after the search loop when `dots == 0` and `RES_NOTLDQUERY` is unset, whatever the loop's last rcode was. So `sudo` was already fine; the client that stopped is c-ares (and it is the reason `_gateway.<search>` got NXDOMAIN in #128). It resolves now too.

For the record, the previous review asked for the opposite ("`runner-abc.lan` must stay on the ordinary path"). I took this one: consistency with `_gateway` and a working c-ares outweigh the shadowing risk, which is the same one already accepted for `_gateway.<search>`, and under filtering the ordinary path refused that form anyway.

pkg/dns/synthetic.go
matthewdevenny: Fixed in c22ba83. `machineNames()` yields nothing for a hostname in the localhost family, so `localhost.local` can never be minted; `TestMachineNames` covers `localhost` and `build.localhost`. (The `.local` form is gone altogether — see the thread below.)

pkg/dns/synthetic.go
matthewdevenny: Fixed in c22ba83. `_localdnsstub` and `_localdnsproxy` are relay-only now (`relayOnlyNames`): answered, never tracked, never written — the same footing as the localhost family, for the reason you give. `TestServeSynthetic_StubListenersUntracked` pins that a rule naming `_localdnsstub` relays `127.0.0.53` while `ipToHostname` and `hostnameIPs` stay empty and the firewall is untouched.

pkg/dns/synthetic.go
matthewdevenny: Declined, with reasoning. Without cargowall the host behaves the same way: resolved synthesizes the configured hostname *before* any upstream lookup, so a workload resolving its own FQDN through the stub already gets the local address, not the VIP or registry record — the relay reproduces what the host did before we arrived, and `s.upstream` was never going to see that query on a resolved host. (Verified on the box: `getent hosts $(hostname)` comes back `Data from: synthetic` with no `/etc/hosts` entry.) The deny case is the same trade-off as `_gateway`: the name resolves, the connection is judged, and a deny rule writes the deny — resolution ahead of the gate is what makes the block attributable instead of a REFUSED query. On caching, agreed it is a stub round trip per lookup; a TTL-0 answer would be "never expires" in `lruCache`, and resolved charges the same round trip natively.

pkg/dns/synthetic.go
matthewdevenny: Fixed in c22ba83. The hostname is read per lookup — `gethostname` is a trivial syscall, no lock, and it runs after the cheaper string checks — so there is no seed to go stale. `TestServeSynthetic_MachineHostnameRelayedAndTracked` pins a rename mid-run: the new name is served, the old one takes the ordinary path.

pkg/dns/synthetic.go
matthewdevenny: Fixed in c22ba83: the guessed `<label>.local` form is gone. `runner-abc.local` takes the ordinary path (pinned). You are right that it was not local state — resolved's rename-on-conflict alone makes the concatenation wrong, and an unsynthesized `.local` query being multicast to the LAN is exactly the answer this path must never enforce.

README.md
matthewdevenny: Fixed in c22ba83. The README paragraph now says the rule needs systemd-resolved on the runner (the Ubuntu images have it), that the name is answered by resolved's stub, and that without one it does not resolve and the rule opens nothing — keep a CIDR on such a host.

pkg/dns/server.go
matthewdevenny: Taken in c22ba83: `enforceDNSResponse` takes an `answerSource` — `wireAnswer` at the upstream and CNAME pre-resolve callers, `localAnswer` at the relay — so a call site reads as what it is rather than a bare `true`. I kept the decision at the callers rather than hoisting `RecordForwardResolution` out, because the evidence loop sits inside the bypass-only gate and moving it would either widen evidence to bypass names or duplicate the gate; the named type gets the legibility without that.

cmd/prepopulate.go
```diff
@@ -0,0 +1,42 @@
...
+// the same evidence class as the proxy's own answers, and what lets an IP be
+// L7-scoped at all (RegisterL7Identity scopes iff bound) — is minted only for
+// a name the wire answered. A name the proxy itself answers from resolved's
+// local state (dns.Server.LocalAlias: _gateway, the machine's own name) is
+// an address alias no peer presents; scoping its address against it would
+// make every flow there an L7 miss.
+func recordSystemCacheAnswer(configMgr *config.Manager, dnsServer *dns.Server, hostname string, ips []string) {
+	alias := dnsServer.LocalAlias(hostname)
+	for _, ip := range ips {
+		configMgr.UpdateDNSMapping(hostname, ip)
```
jordanjennings: Non-blocking. `LocalAlias` collapses both alias classes into one bool, so this still runs `UpdateDNSMapping` for `relayOnlyNames`.

A rule naming `_localdnsstub` puts `ipToHostname["127.0.0.53"]` into the map; `ApplyRulesToTrackedHostnames` unions `GetIPToHostnameMap()` into its replay set (`pkg/dns/server.go:312-318`), `MatchHostnameRule` matches the rule, and `applyVerdictSide` writes BPF entries for the stub address — the one the proxy's own relay and cache peek depend on.

That contradicts the invariant asserted at `pkg/dns/synthetic.go:48` ("relayed, never tracked, never written") and the one `TestServeSynthetic_StubListenersUntracked` pins, which covers only the `serveSynthetic` path. The behavior predates this PR — `cacheResolver` has always queried `127.0.0.53` directly, bypassing the proxy — but this is the change that states the invariant and adds `LocalAlias` as the mechanism for exactly this class of leak.

Note the fix isn't just "skip `UpdateDNSMapping` when `LocalAlias`": attribution mapping is wanted for `_gateway`, so it needs to distinguish `enforcedNames` from `relayOnlyNames`.

jordanjennings (approved):
One non-blocking item inline on `cmd/prepopulate.go`.

Separately: the machine FQDN interception — a hostname that is also a real DNS name now resolves to resolved's synthesized address rather than the upstream record — is unchanged and isn't recorded as a residual in design.md. It's inherent to making the machine hostname resolve at all, so I'm noting it rather than asking for a change.

cmd/prepopulate.go
matthewdevenny: Fixed in 3801088. `LocalAlias` is replaced by `ClassifyAlias` (`NotAlias` / `TrackedAlias` / `UntrackedAlias`), and `recordSystemCacheAnswer` now records a loopback listener or localhost alias nowhere — no `UpdateDNSMapping`, so nothing for `ApplyRulesToTrackedHostnames` to union in and write — while an enforced alias keeps its attribution mapping and a wire name keeps both. `TestRecordSystemCacheAnswer_WithholdsEvidenceForLocalAliases` now covers all three classes, including `_localdnsstub` → `127.0.0.53` never mapped. Agreed the underlying path predates the PR; this is the change that stated the invariant, so it should hold it. The FQDN-interception residual is recorded in design.md's synthetic-names bullet, with the note that it is the host's native behaviour too.

PR #133, opened 2026-09-17
#102 write the downgrade record before the failure sentinel

### LINKED ISSUE #102: Write the downgrade record before the failure sentinel
From the post-merge review of cargowall-action#72 (item 1) — the action classifies a failure sentinel as policy lockdown vs. generic startup failure using the downgrade record, but \`handlePolicyFetchFailure\`'s fail branch writes the **sentinel first, then the sidecar**:

\`\`\`go
if cmd.FailureFile != "" {
    writeFailureSentinel(cmd.FailureFile, reason)   // consumer may act HERE
}
writeDowngradeFile(downgradeRecord(...LOCKDOWN...), logger)
\`\`\`

A consumer that reads between the two writes sees a lockdown sentinel with no downgrade record and misclassifies it as a generic crash — for the action (default \`fail-on-unsupported: false\`) that meant restoring resolv.conf out from under the still-running lockdown and returning success. The action now mitigates with a settle delay plus a sentinel-text fallback (cargowall-action#73), but text matching is a crutch.

**Fix:** invert the order — write the downgrade record first, sentinel last. The sentinel is the signal consumers poll for, so it should be the *commit* of the state, published only once everything describing that state exists. Same consideration applies to any future sidecar keyed off the sentinel.

Worth a small test asserting the downgrade file exists at the moment the sentinel appears (e.g. wait for the sentinel in a goroutine and read the sidecar immediately).

brinkercode:
## Summary

Under `--api-failure-mode=fail`, `handlePolicyFetchFailure` published the failure sentinel before the downgrade record. The action classifies a sentinel as lockdown from that record, so a reader between the two writes saw a lockdown with no record and treated it as a generic startup crash. With the default `fail-on-unsupported: false`, that restored resolv.conf under a live lockdown and passed the step. The record is now written first and the sentinel last.

## Changes

- `cmd/start.go`: `writeDowngradeFile` runs before `writeFailureSentinel` in the lockdown branch. The `SEMANTICS` comment is unchanged; a new `ORDER` comment records why the sequence matters. `writeDowngradeFile` stays best-effort, so a failed record write still leaves the sentinel written.
- `cmd/start_test.go`: `TestLoadCIConfig_LockdownPublishesSentinelLast`.

## Why this shape

The sentinel is what consumers poll for, so it should be the commit of the state: published only once everything describing that state exists. That fixes the race at the source instead of leaning on the settle delay and sentinel-text fallback from code-cargo/cargowall-action#73, which can stay as a second line of defense.

The test points `downgradeFile` and `FailureFile` at the same path. Both writes atomically replace it, so whichever lands last is what remains, and the test asserts that is the sentinel (`pid=`). That checks the order deterministically, without a goroutine racing two writes that land microseconds apart.

## Test plan

| | `main` (750c22f) | this branch |
|---|---|---|
| `TestLoadCIConfig_LockdownPublishesSentinelLast` | fails: file holds the downgrade record | passes |
| `go test ./...` | | green |
| `go test -race ./cmd/` | | green |
| staticcheck, gofumpt, goimports-reviser | | clean |

**Root BPF tests** (Multipass VMs, one package per run): `pkg/tc`, `pkg/network`, `pkg/steps` pass on 5.15, 6.8 and 7.0. `bpf` and `pkg/origin` pass on 6.8 apart from the known VM-kernel `TestTcEgress/Truncated_IP{,v6}_header` EINVALs. This change touches only `cmd`, which none of those packages import.

## Related

Closes #102. Raised as item 1 of the post-merge review on code-cargo/cargowall-action#72; mitigated action-side in code-cargo/cargowall-action#73.

## Out of scope

- `cg_origin_egress` exceeds the verifier's 1M-instruction limit on stock 5.15 and 7.0 kernels (reproduced at #109 and on `main`, including a real runner job on 7.0), so `bpf`/`pkg/origin` fail there and the cgroup hook falls back to TC-only. The budget in #118 / `design-l7.md` is measured on CI's 6.17-azure, so this sits outside that measurement. Unrelated to this change; happy to open an issue with the kernel matrix.


matthewdevenny (approved):
lgtm

brinkercode:
Thanks for the merge! Happy to take on more low-level bugs or enhancements whenever you've got them. I can also open that verifier issue if it's useful.

